PR #15, opened 2023-08-22
Implicit branch creation

maxmynter:
Allow implicit branch creation when the specified branch does not exist. 

[Related Issue](https://github.com/appliedAI-Initiative/lakefs-spec/issues/13).  

- [x] Add a boolean switch (e.g. create_branch_ok) to the LakeFSFileSystem constructor, with documentation. [Code](https://github.com/appliedAI-Initiative/lakefs-spec/pull/15/files#diff-f8235da430cb8c0f32975a91dcd263e02044b2621322a7e6c97182b53649ea1eR122)
- [x] Add the switch to the LakeFSFileSystem.scope() context manager, same as the other two options. [Code](https://github.com/appliedAI-Initiative/lakefs-spec/pull/15/files#diff-f8235da430cb8c0f32975a91dcd263e02044b2621322a7e6c97182b53649ea1eR173)
- [x] Create the branch via client.branches_api.create_branch before an additive operation if create_branch_ok is True - either by creating if it does not exist (incurs a branch listing, which might be expensive for repos with many branches) OR by branch-ensuring via an unconditional create (needs client-side error handling). [Code](https://github.com/appliedAI-Initiative/lakefs-spec/pull/15/files#diff-f8235da430cb8c0f32975a91dcd263e02044b2621322a7e6c97182b53649ea1eR94)
- [x] Create the branch in the same way for the LakeFSFile if it is opened in write mode, e.g. in __init__ if mode == "wb" OR in _upload_chunk directly before the put_object call). [Code](https://github.com/appliedAI-Initiative/lakefs-spec/pull/15/files#diff-f8235da430cb8c0f32975a91dcd263e02044b2621322a7e6c97182b53649ea1eR393)


maxmynter:
In making this PR, i found two issues. 

1. When using `LakeFSFileSystem.put()` the prefix `lakefs://` in the resource identification string throws an error, e.g. : `ValueError: expected path with structure <repo>/<ref>/<resource>, got 'lakefs://quickstart/main/lakesHead.csv'`. This is not the case for the `get()` method. I think we should support it with the prefix. I am unsure as to what causes this deviant behaviour as both methods use the same `parse()` function

2. Pushing changes with `postcommit = True` results in an error when no changes are on the branch. The error: 

```Reason: Bad Request
HTTP response headers: HTTPHeaderDict({'Content-Type': 'application/json', 'X-Content-Type-Options': 'nosniff', 'X-Request-Id': '490cc54c-a8f9-41c4-974b-25e56c9b9aa3', 'Date': 'Tue, 22 Aug 2023 13:54:27 GMT', 'Content-Length': '33'})
HTTP response body: {"message":"commit: no changes"}
```

I think we should handle that error since we also have a logging statement informing that there was no update to the resource.

src/lakefs_spec/spec.py
```diff
@@ -352,21 +370,29 @@ def put(
...
     ):
+        repository, branch, resource = parse(rpath)
+
         super().put(
             lpath, rpath, recursive=recursive, callback=callback, maxdepth=maxdepth, **kwargs
         )
 
+        if self.create_branch_ok:
+            check_branch_exists_or_create(self.client,repository, branch)
+
```
nicholasjng, line 381: This should probably happen before the call to `put` - and if you use `parse` there, it should (as of now) come with a `rpath = self._strip_protocol(rpath)` before. (But we should change the parsing so that is not required.)

src/lakefs_spec/spec.py
```diff
@@ -90,6 +91,17 @@ def parse(path: str) -> tuple[str, str, str]:
     repo, ref, resource = results.groups()
     return repo, ref, resource
 
+def check_branch_exists_or_create(client, repository, new_branch_name:str, source_branch_name:str = 'main'):
```
nicholasjng, line 94: What do you think about a name like `ensure_branch`? Since in this way, we _ensure_ that the branch exists after this code executes, either by creating it or passing on error. 

There is the `ensurerepo` function in the tests as precedent - maybe this should be named in the same manner.

maxmynter: Like it :)

src/lakefs_spec/spec.py
```diff
@@ -352,21 +370,29 @@ def put(
...
                 repository=repository, branch=branch, commit_creation=commit_creation
             )
+        
 
     def rm_file(self, path):
         repository, branch, resource = parse(path)
 
+        if self.create_branch_ok:
+            check_branch_exists_or_create(self.client, repository, branch)
+
```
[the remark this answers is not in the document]
maxmynter: Agree

src/lakefs_spec/spec.py
```diff
@@ -90,6 +91,17 @@ def parse(path: str) -> tuple[str, str, str]:
     repo, ref, resource = results.groups()
     return repo, ref, resource
 
+def check_branch_exists_or_create(client, repository, new_branch_name:str, source_branch_name:str = 'main'):
+        try:
+            #TODO: (m.mynter) How to get the source branch in implicit branch creation?
```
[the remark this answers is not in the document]
maxmynter: Yes, I think it makes sense to have it there. AFAIK there is no possibility to get the previously checked out branch for lakeFS. Plus, maybe I want another branch as source branch.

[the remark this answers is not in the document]
nicholasjng: Makes sense. So between

- memoizing the last checked out branch for each resource (presumably in a cache?),
- giving the option of a default file system-wide base branch,
- and always branching off of main,

option 2) is the best IMO. Making `main` the default is probably good enough for 90% of cases, but setting it as an attribute is good practice anyway.

What do you think? If you want to go ahead without, it can also go into a follow-up.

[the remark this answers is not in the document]
maxmynter: - [Change](https://github.com/appliedAI-Initiative/lakefs-spec/pull/15/files#diff-f8235da430cb8c0f32975a91dcd263e02044b2621322a7e6c97182b53649ea1eR457) for `LakeFSFile` 
- - [Change](https://github.com/appliedAI-Initiative/lakefs-spec/pull/15/files#diff-f8235da430cb8c0f32975a91dcd263e02044b2621322a7e6c97182b53649ea1eR143) for `LakeFSFileSystem`

src/lakefs_spec/spec.py
```diff
@@ -127,12 +161,18 @@ def __init__(
         precheck_files: bool
             Whether to compare MD5 checksums of local and remote objects before file
             operations, and skip these operations if checksums match.
+        create_branch_ok: bool
+            Whether to create branches implicitly when not-existing branches are referenced.
```
nicholasjng, line 167:
```suggestion
            Whether to create branches implicitly when not-existing branches are referenced on file uploads.
```

src/lakefs_spec/spec.py
maxmynter: Here I am unsure whether or not I adressed your comment with [these Changes](https://github.com/appliedAI-Initiative/lakefs-spec/pull/15/files#diff-f8235da430cb8c0f32975a91dcd263e02044b2621322a7e6c97182b53649ea1eR414). If not, please elaborate :)

src/lakefs_spec/spec.py
```diff
@@ -91,6 +91,38 @@ def parse(path: str) -> tuple[str, str, str]:
     return repo, ref, resource
 
 
+def ensure_branch(
+    client: LakeFSClient, repository: str, branch_name: str, source_branch_name: str = "main"
```
nicholasjng, line 97: Could you rename these arguments to `branch` and `source_branch`, respectively? Same in the docstrings.

src/lakefs_spec/spec.py
```diff
@@ -91,6 +91,38 @@ def parse(path: str) -> tuple[str, str, str]:
...
+    branch_name: str
+        Name of the branch.
+    source_branch_name: str
+        Name of the source branch the new branch is created from.
+
+    Returns
+    -------
+    None
+    """
+    client.commits
```
nicholasjng, line 118: This is probably an accidental leftover from development?

src/lakefs_spec/spec.py
```diff
@@ -329,6 +384,9 @@ def put_file(
     ):
         repository, branch, resource = parse(rpath)
 
+        if self.create_branch_ok:
+            ensure_branch(self.client, repository, branch, self.source_branch)
+
```
nicholasjng, line 389: This will probably result in endpoint thrashing if you upload 2000 data frame shards.

Should we agree on only ensuring the branch once in `fs.put()`? After all, `put_file` is more of a lower-level method, that should probably not be called directly.

src/lakefs_spec/spec.py
```diff
@@ -398,6 +458,7 @@ def __init__(
         fs,
         path,
         mode="rb",
+        source_branch="main",
```
nicholasjng, line 461: This is not needed in the file constructor, since you can get it out of the file system as `self.fs.source_branch`.

src/lakefs_spec/spec.py
```diff
@@ -424,6 +485,9 @@ def __init__(
             size=size,
             **kwargs,
         )
+        if mode == "wb":
+            repository, branch, resource = parse(path)
+            ensure_branch(self.fs.client, repository, branch, source_branch)
```
nicholasjng, line 490:
```suggestion
            ensure_branch(self.fs.client, repository, branch, self.fs.source_branch)
```

src/lakefs_spec/spec.py
```diff
@@ -424,6 +481,9 @@ def __init__(
             size=size,
             **kwargs,
         )
+        if mode == "wb":
+            repository, branch, resource = parse(path)
+            ensure_branch(self.fs.client, repository, branch, self.fs.source_branch)
```
nicholasjng, line 486: Can we merge this with the warning above the `super()` call? Just move the warning here and delete the first if-branch.

src/lakefs_spec/spec.py
```diff
@@ -93,6 +93,38 @@ def parse(path: str) -> tuple[str, str, str]:
     return repo, ref, resource
 
 
+def ensure_branch(
+    client: LakeFSClient, repository: str, branch: str, source_branch: str = "main"
```
nicholasjng, line 97:
```suggestion
    client: LakeFSClient, repository: str, branch: str, source_branch: str
```

Please drop the default, otherwise this might silently work when it really shouldn't

nicholasjng (changes requested):
Let's add a follow-up issue for a `create_branch_ok` test (and, more importantly, a "create branch not ok" failure) - otherwise, we're set!

maxmynter:
> Let's add a follow-up issue for a create_branch_ok test (and, more importantly, a "create branch not ok" failure) - otherwise, we're set!

New issue created [here](https://github.com/appliedAI-Initiative/lakefs-spec/issues/23)

nicholasjng (approved):

PR #18, opened 2023-08-23
Remove `lakeFS` client stub

### LINKED ISSUE #17: Clean up `LakeFSClient` stub
#9 bumped the minimum lakeFS client version to `v0.105.0`, which means that typing information is now available for the client. Thus, the `client.{py,pyi}` files can go.

Also in that release came the deprecation of client API members without `_api` suffixes, so we should migrate the usage of client APIs to the `_api` suffix variants.

nicholasjng:
The proper typing information was upstreamed and included in v0.105.0.

Since that is the minimum we require now, we can safely delete the stub file again and rely on the package typing info.

Also switch out API member calls on the lakeFS client.

The non-suffixed members, e.g. `client.objects` or `client.branches`, were deprecated in v0.105.0, so we switch out to the suffixed members instead.

These were always the same anyways (matching object IDs), so there is no change in behavior.

Fixes #17.


nicholasjng:
cc @AdrianoKF

PR #19, opened 2023-08-23
PyPI publishing

AdrianoKF:


.github/workflows/python.yaml
```diff
@@ -2,6 +2,8 @@ name: lakefs-spec CI
 
 on:
   push:
+    tags:
+      - "v*"
```
nicholasjng, line 6: As per our discussion, I would propose making this a different trigger based on release cuts on GH like so:

```yaml

on:
  workflow_dispatch:
  release:
    types:
      - published
```

src/lakefs_spec/__init__.py
```diff
@@ -1 +1 @@
-version = "0.1.0"
+version = "0.1.1"
```
nicholasjng, line 1: Why is this not dynamic? Seems like the ideal opportunity to just grab the version from here in the `pyproject.toml`.

To do this, add 

```
dynamic = ["version"]
```
to the `[project]` block, and 

```
[tool.setuptools.dynamic]
version = { attr = "lakefs_spec.version" }
```
somewhere below.

nicholasjng (changes requested):
Have you already pushed `v0.1.0` to real PyPI now?

The trigger setup is the only obligatory change here in my opinion, since the PyPI publish job should not appear in the normal pull request actions tab IMO.

src/lakefs_spec/__init__.py
AdrianoKF: I was going to create a follow-up issue to address version bookkeeping (e.g., with bumpversion), but we can also make the change here.

.github/workflows/python.yaml
```diff
@@ -31,3 +33,36 @@ jobs:
...
+    - name: Upload build artifacts
+      uses: actions/upload-artifact@v3
+      with:
+        name: dist
+        path: dist
+
+  publish_pypi:
+    name: Publish wheels to PyPI
+    needs: [build]
+    runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
AdrianoKF: Probably yes, I believe `on` is per-file.

nicholasjng (approved):

PR #21, opened 2023-08-23
Switch lakeFS CI instance to service container

nicholasjng:
https://github.com/treeverse/lakeFS/issues/6452 revealed that there is a possibility to emulate the quickstart via some environment variables.

This commit sets them and defines a GitHub Actions service container from the latest lakeFS tag to try and see if that works.

### DIFF
--- a/.github/workflows/python.yaml
+++ b/.github/workflows/python.yaml
@@ -16,6 +16,18 @@ permissions:
 jobs:
   test:
     runs-on: ubuntu-latest
+    services:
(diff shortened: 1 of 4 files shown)


nicholasjng:
Just as a side note: This change is not 100% equivalent, since we lose the dynamic tag from the `lakefs_client.__version__`. 
This ensured that client and server are tested in lockstep. But the service container here seems like a big enough improvement that we can keep them in sync by hand.

Issue #2, opened 2023-08-10
Set up GitHub Actions for the project


nicholasjng:
Additionally to the existing CI setup, we would need a lakeFS instance for integration tests.

Subtasks:

- [x] Add `pre-commit` job
    - [x] Primed on pull requests against `main` and merges into `main`
    - [x] Bonus: with caching via `actions/cache` (including tool subcaches)
- [x] Add `pytest` job
    - [x] Primed on pull requests against `main` and merges into `main`
    - [x] ~~Add a `service` block for `lakeFS` setup (local mode should be fine)~~
    - [x] Bonus: with venv caching via `actions/cache` (keyed on `dev-deps.lock`)

nicholasjng:
Closed via #24.

PR #24, opened 2023-08-24
feat(ci): Add Python venv cache

AdrianoKF:
Issue: #2

### DIFF
--- a/.github/workflows/python.yaml
+++ b/.github/workflows/python.yaml
@@ -26,6 +26,9 @@ jobs:
           LAKEFS_DATABASE_TYPE: "local"
           LAKEFS_AUTH_ENCRYPT_SECRET_KEY: "THIS_MUST_BE_CHANGED_IN_PRODUCTION"
           LAKEFS_BLOCKSTORE_TYPE: "local"
+    env:
+      MYPY_CACHE_DIR: "${{ github.workspace }}/.cache/mypy"
+      RUFF_CACHE_DIR: "${{ github.workspace }}/.cache/ruff"    
     steps:
     - uses: actions/checkout@v3
     - name: Set up Python 3.11
@@ -45,12 +48,12 @@ jobs:
       uses: actions/cache@v3
       with:
         path: |
-          ~/.cache/ruff/
-          ~/.cache/mypy/
+          ${{ env.MYPY_CACHE_DIR }}
+          ${{ env.RUFF_CACHE_DIR }}
         key: ${{ runner.os }}-${{ hashFiles('dev-deps.lock', 'pyproject.toml', '.pre-commit-config.yaml') }}-linter-cache
     - uses: pre-commit/action@v3.0.0
       with:
-        extra_args: --show-diff-on-failure
+        extra_args: --all-files
     - name: Test with pytest
       run: |
         pytest -s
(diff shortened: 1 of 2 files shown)


.github/workflows/python.yaml
```diff
@@ -31,17 +29,28 @@ jobs:
...
-    - name: Lint with pre-commit
-      run: |
-        pre-commit run -av
+    - name: Cache linter tool caches
+      uses: actions/cache@v3
+      with:
+        path: |
+          ~/.cache/ruff/
+          ~/.cache/mypy/
+        key: ${{ runner.os }}-${{ hashFiles('dev-deps.lock', 'pyproject.toml', '.pre-commit-config.yaml') }}-linter-cache
```
nicholasjng, line 50: Why `dev-deps.lock` and `pyproject.toml`? For caching on the pre-commit version, too?

.github/workflows/python.yaml
```diff
@@ -31,17 +29,28 @@ jobs:
...
+    - uses: pre-commit/action@v3.0.0
+      with:
+        extra_args: --show-diff-on-failure
```
nicholasjng, line 53: The action already supplies `--show-diff-on-failure`, see https://github.com/pre-commit/action/blob/5f528da5c95691c4cf42ff76a4d10854b62cbb82/action.yml#L19

.github/workflows/python.yaml
```diff
@@ -31,17 +29,28 @@ jobs:
...
         pip install -e . --no-deps
-    - name: Lint with pre-commit
-      run: |
-        pre-commit run -av
+    - name: Cache linter tool caches
+      uses: actions/cache@v3
+      with:
+        path: |
+          ~/.cache/ruff/
+          ~/.cache/mypy/
```
nicholasjng, line 49: I think these tools open caches in the working directory by default, so we need to set some envvars for changing the cache location - they were in the Gitlab CI, let me take a look.

nicholasjng (changes requested):

.github/workflows/python.yaml
nicholasjng: This is from the gitlab CI:

```yaml
  PRE_COMMIT_HOME: "$CI_PROJECT_DIR/.cache/pre-commit"
  MYPY_CACHE_DIR: "$CI_PROJECT_DIR/.cache/mypy"
  RUFF_CACHE_DIR: "$CI_PROJECT_DIR/.cache/ruff"
```

Since pre-commit is probably already primed on `$HOME` (I assume, from the action), substituting `~` as `$CI_PROJECT_DIR` should do the trick?

.github/workflows/python.yaml
nicholasjng: I'm not really sure about `pyproject.toml` - it only contains a "greater-or-equal" pin for `pre-commit`, and hashing it will cause cache misses on option changes for all tools (and unrelated metadata changes, too). The relevant version data seems to be contained entirely in `dev-deps.lock`.

What do you think?

AdrianoKF: Fine with me, I just like to err on the side of caution when it comes to caching. Will change

nicholasjng (approved):

PR #29, opened 2023-08-24
Change lakeFS file system instantiation mechanism to take raw arguments

### LINKED ISSUE #16: Pass configuration arguments instead of client to the `LakeFSFileSystem` constructor
This is similar to, for example, how `s3fs` does it: https://github.com/fsspec/s3fs/blob/39125f79051c55fc715cf056d8a2d0c7aa9d1c4b/s3fs/core.py#L171

It also has another benefit: Giving the bare options instead of the client should lead to more robust instance caching, since the cache key is now calculated from the config attributes directly and not the client.

### Implementation

- [ ] Pick out `lakefs_client.Configuration` attributes that should be supported.
- [ ] Add them to the `LakeFSFileSystem` constructor.
- [ ] Change all example code to take the raw config attributes as `storage_options`.

nicholasjng:
Instead of an instantiated client, take the raw lakeFS configuration arguments to construct the lakeFS file system. This has the distinct advantage that no `lakefs_client` imports need to happen on the user's part.

For e.g. `pandas` usage, this has the effect that all attributes (e.g. host, username) can be given directly into the `storage_options` dict. The readme was updated to showcase this.

Also, instead of a direct config, environment variables were defined, which can be set to use zero-code config for the file system, and thus skip `storage_options` altogether.

All tests were updated to use the new mode of instantiation. In the future, we can switch to envvar-config and drop the fixture configuration entirely (except for a unit test verifying correct behavior).

Closes #16.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,14 +39,15 @@ Then, run the following code to download the sample dataframe directly from the 
 
 ```python
 import pandas as pd
-from lakefs_client import Configuration
-from lakefs_client.client import LakeFSClient
 
-# change these settings to match your instance's credentials
-configuration = Configuration(host="localhost:8000", username="username", password="password")
-client = LakeFSClient(configuration=configuration)
+# change these settings to match your instance's address and credentials
+storage_options={
+    "host": "localhost:8000",
+    "username": "username",
+    "password": "password",
+}
 
-df = pd.read_parquet('lakefs://quickstart/main/lakes.parquet', storage_options={"client": client})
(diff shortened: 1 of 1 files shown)


AdrianoKF:
IMO, a few additional configuration options could be exposed:

- `verify_ssl` (which unfortunately is not part of `Configuration`, but rather the lakeFS client itself - but for user convenience, I would make it part of our constructor)
- `proxy` (same argument)
- `ssl_ca_cert` (needed for self-managed PKI)

PR #32, opened 2023-08-25
Simplify tests

AdrianoKF:
This PR simplifies most tests by introducing a new fixture for the `LakeFSFileSystem` instance under test.

It also includes a docker-compose file for local development, which creates an ephemeral lakeFS quickstart instance with settings compatible to the defaults in the test fixtures.

### DIFF
--- a/hack/docker-compose.yml
+++ b/hack/docker-compose.yml
@@ -1,4 +1,4 @@
-# This docker-compose file provides an ephemeral lakeFS quickstart instance for local development
+# This docker-compose file provides an ephemeral lakeFS quickstart instance for local development.
 
 version: "3"
 
--- a/hack/docker-compose.yml
+++ b/hack/docker-compose.yml
@@ -1,0 +1,16 @@
+# This docker-compose file provides an ephemeral lakeFS quickstart instance for local development
+
+version: "3"
+
+services:
+  lakefs:
+    image: treeverse/lakefs:0.107.0
+    ports:
+      - 8000:8000
+    environment:
+      LAKEFS_INSTALLATION_USER_NAME: "quickstart"
+      LAKEFS_INSTALLATION_ACCESS_KEY_ID: "AKIAIOSFOLQUICKSTART"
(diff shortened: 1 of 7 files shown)


tests/conftest.py
```diff
@@ -11,6 +11,7 @@
 from lakefs_client.client import LakeFSClient
 from lakefs_client.models import BranchCreation, RepositoryCreation
 
+from lakefs_spec.spec import LakeFSFileSystem
```
nicholasjng, line 14:
```suggestion
from lakefs_spec import LakeFSFileSystem
```

Since #30 this is now `from lakefs_spec import LakeFSFileSystem`.

hack/docker-compose.yml
```diff
@@ -0,0 +1,16 @@
+# This docker-compose file provides an ephemeral lakeFS quickstart instance for local development
```
nicholasjng, line 1:
```suggestion
# This docker-compose file provides an ephemeral lakeFS quickstart instance for local development.
```

tests/conftest.py
```diff
@@ -40,6 +41,11 @@ def lakefs_options() -> LakeFSOptions:
     )
 
 
+@pytest.fixture
+def fs(lakefs_options: LakeFSOptions) -> LakeFSFileSystem:
+    return LakeFSFileSystem(**lakefs_options._asdict())
```
nicholasjng, line 46: Might have been better on my part to make it a dataclass and use `asdict` here to save an underscored access, oh well.

hack/docker-compose.yml
AdrianoKF: Addressed in 5c3d306

tests/conftest.py
AdrianoKF: Addressed in 4e16c07b50d4051f3e9b4563b1a9bc8142beb6c5

tests/conftest.py
AdrianoKF: Made `LakeFSOptions` a dataclass in 4e16c07b50d4051f3e9b4563b1a9bc8142beb6c5

PR #36, opened 2023-08-28
Improve flexibility of commithooks by adding more arguments

### LINKED ISSUE #27: Improve commit hook abstraction
Right now, we only supply the `fsspec` event name, and the resource that is being mutated.

However, to craft a better commit message automatically, more information would be nice. This could include:

- The branch name that is being committed to.
- The name of the repository holding the branch.
- The diff that is being committed.

We already have all the above information present (with the changes to the commit hook logic in #26), so this should not be hard to do - only the interface change needs to be implemented (and communicated!)

### Implementation
- [ ] Decide what of the above info to include for commit hooks.
- [ ] Updating the `CommitHook` type hint in the `src/lakefs_spec/commithook.py` file.
- [ ] Updating the `Default` commit hook to take the new arguments.
- [ ] Updating the commit hook unit test.

nicholasjng:
Adds the repository, the branch, and the diff that is being committed as direct inputs to the commit hook.

Also updates the docstring for the `CommitHook` type. Arguments were renamed from `event` to `fs_event`, and `rpath` to `resource`, respectively, to avoid confusion.

Fixes #27.

### DIFF
--- a/src/lakefs_spec/commithook.py
+++ b/src/lakefs_spec/commithook.py
@@ -1,29 +1,42 @@
-from typing import Callable
+from enum import Enum
+from typing import Callable, NamedTuple
 
-from lakefs_client.models import CommitCreation
+from lakefs_client.models import CommitCreation, DiffList
 
-CommitHook = Callable[[str, str], CommitCreation]
+
+class FSEvent(Enum):
+    PUT = 1
+    RM = 2
+
+
+class HookContext(NamedTuple):
+    repository: str
+    branch: str
+    resource: str
+    diff: DiffList
+
+
+CommitHook = Callable[[FSEvent, HookContext], CommitCreation]
 """
-A hook to execute before a lakeFS commit is created during stateful file operations such as
-uploads or deletes. Input arguments are fsspec event name (e.g. ``put``, ``rm``) and rpath,
-the output needs to be a ``lakefs_client.models.CommitCreation`` object to pass to the
-``LakeFSClient.commits.commit`` API.
+A hook to execute before a lakeFS commit is created during stateful file operations such as uploads or deletes.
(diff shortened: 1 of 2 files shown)


nicholasjng:
> A few design comments:
> 
>     * I would introduce a union type or enum for the events, to make typing easier for developers adding their own hooks.
> 
>     * Instead of passing all the hook data as args, I would wrap them in a `HookContext` dataclass, so that the interface for a hook is stable.
> 
> 
> Combining these, the signature for a hook function would look like this: `(EventType, HookContext) -> CommitCreation`

Both great suggestions, I added them in 4432335.

src/lakefs_spec/commithook.py
```diff
@@ -35,7 +35,7 @@ def Default(fs_event: FSEvent, ctx: HookContext) -> CommitCreation:
     elif fs_event == FSEvent.RM:
         action = "Remove"
     else:
-        raise ValueError(f"unexpected file system event {fs_event!r}")
+        raise ValueError(f"unexpected file system event {str(fs_event)!r}")
```
AdrianoKF, line 38: Maybe just drop the `!r` flag? `f"... {fs_event}"` should work fine.

nicholasjng: That's needed for the quotes around the event name - which I want to make clear as the input to the hook in the error:

`ValueError: Unexpected file system event 'foobar'` vs. `ValueError: Unexpected file system event foobar`

I'm trying to do the former. What do you think?

AdrianoKF: Phew. I'd just go with f'..."{fs_event}"', which makes it explicit that you're after the quotes.

nicholasjng: Done in a5206aa.

PR #37, opened 2023-08-28
Add more configuration arguments to lakeFS file system

### LINKED ISSUE #31: Expose additional lakeFS client configuration attributes in constructor
> IMO, a few additional configuration options could be exposed:
>
>- `verify_ssl` (which unfortunately is not part of `Configuration`, but rather the lakeFS client itself - but for user convenience, I would make it part of our constructor)
>- `proxy` (same argument)
>- `ssl_ca_cert` (needed for self-managed PKI)

_Originally posted by @AdrianoKF in https://github.com/appliedAI-Initiative/lakefs-spec/issues/29#issuecomment-1692896610_

nicholasjng:
The three additional arguments added here are:

- `verify_ssl`, to toggle whether or not to verify SSL certificates in HTTPS connections
- `ssl_ca_cert`, pointing to a PEM file for cert verification
- `proxy`, for giving a proxy server address.

Closes #31.

### DIFF
--- a/src/lakefs_spec/spec.py
+++ b/src/lakefs_spec/spec.py
@@ -143,6 +143,9 @@ class LakeFSFileSystem(AbstractFileSystem):
         api_key: str | None = None,
         api_key_prefix: str | None = None,
         access_token: str | None = None,
+        verify_ssl: bool = True,
+        ssl_ca_cert: str | None = None,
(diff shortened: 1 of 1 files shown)


AdrianoKF:
LGTM

PR #46, opened 2023-08-30
chore: Lock main dependencies for CI

AdrianoKF:
This PR adds a lock file for the main dependencies, `requirements.txt`, obtained through pip-compile, similar to the existing dev dependencies.

The lockfile for dev dependencies has been renamed to `requirements-dev.txt` for consistency.

The contribution guide and CI actions have been updated to reflect the new changes.

Also, this change adds pre-commit hooks for automatically maintaining the lockfiles via `pip-compile`.

⚠️ Due to a breaking API change in `lakefs-client==0.108.0`, the dependency has been pinned at `0.107.1` for now.


.github/workflows/python.yaml
```diff
@@ -38,11 +38,13 @@ jobs:
...
         cache-dependency-path: |
-          dev-deps.lock
+          requirements.txt
+          requirements-dev.txt
           pyproject.toml
     - name: Install dependencies
       run: |
-        pip install -r dev-deps.lock -e .
+        pip install -r requirements.txt -r requirements-dev.txt
+        pip install -e . --no-deps
```
nicholasjng, line 47: Nice! Much more predictable as two commands imo

nicholasjng (commented):
Looks great! Let's see about the new client version, also thank you for adding discussions.

AdrianoKF:
> Looks great! Let's see about the new client version, also thank you for adding discussions.

We could also pin lakeFS to 0.107.0 for the time being and deal with the API change in a follow-up PR for clarity.

nicholasjng (approved):

AdrianoKF:
Follow-up issue for the breaking API change: #47

PR #48, opened 2023-08-30
Add implicit branch creation and other small changes to Readme.md

maxmynter:
As in title. 

Added an explanation of the implicit branch creation and uploading files in general to the `readme.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,10 @@
 # lakefs-spec: An `fsspec` implementation for lakeFS
+
 This repository contains a [filesystem-spec](https://github.com/fsspec/filesystem_spec) implementation for the [lakeFS](https://lakefs.io/) project.
 Its main goal is to facilitate versioned data operations in lakeFS directly from Python code, for example using `pandas`. See the [examples](#usage) below for inspiration.
 
 ## Installation
+
 To install the package directly from PyPI via `pip`, run
 
 ```shell
@@ -28,6 +30,7 @@ poetry add git+https://github.com/appliedAI-Initiative/lakefs-spec.git
 ```
 
 ## Usage
+
 As an example showcase, we use the lakeFS file system to read a Pandas `DataFrame` directly from a branch. To follow
 this small tutorial, you should first complete Step 1 in the [lakeFS quickstart](https://docs.lakefs.io/quickstart/launch.html) by
 launching an instance, and then creating a pre-populated repository by clicking the green button on the login page.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -56,6 +56,27 @@ The lakeFS filesystem expects URIs that follow the [lakeFS protocol](https://doc
...
 
+### Updating files
+
+You can update data in LakeFS like so:
+
+```python
+df.to_csv('lakefs://quickstart/main/lakes.parquet', storage_options=storage_options)
+```
+
+When the target file not exists, it is created. Otherwise, the existing file is updated.
```
nicholasjng, line 67:
```suggestion
If the target file does not exist, it is created, otherwise, the existing file is updated.
```

README.md
```diff
@@ -56,6 +56,27 @@ The lakeFS filesystem expects URIs that follow the [lakeFS protocol](https://doc
...
+
+If the specified branch does not exist, it is created. This behaviour can be set in the filesystem constructor via the `create_branch_ok` flag:
```
nicholasjng, line 69:
```suggestion
If the specified branch does not exist, it is created by default. This behaviour can be set in the filesystem constructor via the `create_branch_ok` flag:
```

README.md
```diff
@@ -56,6 +56,27 @@ The lakeFS filesystem expects URIs that follow the [lakeFS protocol](https://doc
...
+
+```python
+from lakefs_spec import LakeFSFileSystem
+
+# The default setting, precheck_files=False disables client-side caching.
```
nicholasjng, line 74: This comment is stale / copy-pasted?

README.md
```diff
@@ -109,7 +129,7 @@ fs = LakeFSFileSystem(host="localhost:8000", postcommit=True, commithook=my_comm
 
 ### Scoped filesystem behavior changes
 
-To selectively enable or disable automatic commits or client-side caching, you can use a `scope` context manager:
+To selectively enable or disable automatic commits, client-side caching, or implicit branch creation you can use a `scope` context manager:
```
nicholasjng, line 132:
```suggestion
To selectively enable or disable automatic commits, client-side caching, or automatic branch creation, you can use a `scope` context manager:
```

nicholasjng (changes requested):
Could you just append the bit on writing pandas DataFrames to the pandas example? 

Currently, the bit is too specific to pandas to be its own section, since we do not want to assume the user actually using pandas.

README.md
maxmynter: jup

README.md
```diff
@@ -50,6 +50,26 @@ storage_options={
...
+```
+
+If the target file does not exists, it is created, therwise, the existing file is updated.
+
+If the specified branch does not exist, it is created by default. This behaviour can be set in the filesystem constructor via the `create_branch_ok` flag.
+
+```python
+from lakefs_spec import LakeFSFileSystem
+
+# The default setting, precheck_files=True enables implicit branch creation
```
nicholasjng, line 66:
```suggestion
# create_branch_ok=True (the default setting) enables implicit branch creation
```

README.md
```diff
@@ -93,7 +113,6 @@ def my_commit_hook(event: FSEvent, ctx: HookContext) -> CommitCreation:
         message = f"❌ Remove file {ctx.resource}"
     else:
         message = f"✅ Add file {ctx.resource}"
-
```
nicholasjng, line 96: Please revert, the space makes it more readable because of the different indentations (imo).

nicholasjng (changes requested):

PR #64, opened 2023-09-12
Add API call wrapper context manager

nicholasjng:
So far, its arguments are unused since we translate the lakeFS errors raw and without custom messages.

Exchanges the verbose try-excepts for an indent level due to the new context manager.


nicholasjng:
And there we have it boys, 90+% coverage. Not sure how much that's worth, but it's a start

PR #67, opened 2023-09-12
Create a Demo for LakeFS-spec and Using LakeFS in a small scale e-2-e example project

maxmynter:
# Changes
- Additional Environment, `demo-requirements.txt`: The demo needs additional dependencies to the `LakeFS-spec` package, as we want it to be an end to end project. Therefore, I added an additional environment as to not dilute the package dependencies

# Questions to reviewer
- Currently, I am specifiying a selected commit in the `demo-requirements.txt`, `lakefs-spec @ git+https://github.com/appliedAI-Initiative/lakefs-spec.git@e2d188c937310b6053759e7db56804235c64e352` in order to keep the demo intact even if we create breaking changes. Do you agree with this approach?


maxmynter:
Will definetly squash-merge this, once approved :)

demo-requirements.txt
nicholasjng, line 1: This does not belong in the main project - it should go into a `demo/` subfolder (e.g. `rain_prediction` or `e2e`) and live there together with the notebook.

From there, you can also decide to `!pip install -r demo-requirements.txt` directly in the notebook (you can drop the demo prefix once it lives in the demo subfolder mentioned above), assuming the user has his Jupyter Notebook server in a virtual environment already. (At least that's what I have seen in a lot of demos)

.gitignore
```diff
@@ -67,6 +67,7 @@ __pypackages__/
 # Environments
 .env
 .venv
+.demo-venv
```
nicholasjng, line 70: How about sticking with `venv` or `.venv` as a folder name inside `demos/`, and letting gitignore handle it with the already excluded names?

nicholasjng (commented):
Preliminary demo structure comments, will dive into the contents later.

demos/rain_prediction.ipynb
```diff
@@ -0,0 +1,467 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Now, on LakeFS in your browser, can change the branch to `transform-raw-data` and see the saved file. However, the change is not yet committed. \n",
+    "While you can do that manually via the uncommitted changes tab in Lakefs, we will take another route to commit. \n",
+    "\n",
+    "We can also do commit changes programmatically. To achieve this, we register a hook. The hook needs to have the signature `(client, context) -> None`. Where the `client` is the respective LakeFS client. The context object contains information about the corresponding resource. \n",
```
nicholasjng, line 86:
```suggestion
    "To commit changes programmatically, we can register a hook. This hook needs to have the signature `(client, context) -> None`, where the `client` is the file system's LakeFS client. The context object contains information about the requested resource. \n",
```

demos/rain_prediction.ipynb
```diff
@@ -0,0 +1,467 @@
...
+   "source": [
+    "from lakefs_client.client import LakeFSClient\n",
+    "from lakefs_spec.client_helpers import commit\n",
+    "from lakefs_spec.hooks import FSEvent, HookContext\n",
+    "\n",
+    "#Define the commit hook\n",
+    "def commit_on_put(client: LakeFSClient, ctx:HookContext) -> None:\n",
+    "    commit_message = f\"Add file {ctx.resource}\"\n",
+    "    print(f\"Attempting Commit: {commit_message}\")\n",
+    "    commit(client, repository=ctx.repository, branch = ctx.ref, message=commit_message)\n",
```
nicholasjng, line 104:
```suggestion
    "    commit(client, repository=ctx.repository, branch=ctx.ref, message=commit_message)\n",
```

demos/rain_prediction.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "\n",
+    "#Register the commit hook to be executed after a PUT_FILE event\n",
+    "fs.register_hook(FSEvent.PUT_FILE, commit_on_put)\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "When you execute the next cell however, you will see a message indicating that the upload of the resource has been skipped because the file is uploaded to LakeFS already (checksums match). This is useful when we work with large files to reduce the amount of network traffic. Nonetheless, in this specific situation the `PUT` is not executed and neither is our commit hook. \n"
```
nicholasjng, line 115: Whoops, this kind of sucks, right? Defeats the concept of a hook in the first place. Could be worth an issue?

```suggestion
    "However, when you execute the next cell, you will see a message indicating that the upload of the resource has been skipped because the file is uploaded to lakeFS already, and the checksums match. This is useful when we work with large files to reduce the amount of network traffic. Nonetheless, in this specific situation the `PUT` is not executed and neither is our commit hook. \n"
```

demos/rain_prediction.ipynb
```diff
@@ -0,0 +1,467 @@
...
+   "outputs": [],
+   "source": [
+    "fs.put('./data/weather-2010s.json',  f'{REPO_NAME}/{NEW_BRANCH_NAME}/weather-2010.json')"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We can circumvent this by disabling checking the checksums on a specific put operation. We do this by passing `precheck=False` to the `PUT` operation."
```
nicholasjng, line 131:
```suggestion
    "We can circumvent this by disabling pre-checking file checksums on a specific put operation. We do this by passing `precheck=False` to the `PUT` operation."
```

demos/rain_prediction.ipynb
```diff
@@ -0,0 +1,467 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Now let's transform the data for our use case. We put the transformation into a function such that we can reuse it later\n",
+    "\n",
+    "In this notebook, we follow a simple toy example to predict whether it is raining at the same time tomorrow given weather data from right now. \n",
+    "\n",
+    "However, we will skip a lot of possible feature engineering etc. in order to focus on the application of LakeFS and LakeFS-spec. "
```
nicholasjng, line 151:
```suggestion
    "We will skip a lot of possible feature engineering etc. in order to focus on the application of lakeFS and the `LakeFSFileSystem`."
```

demos/rain_prediction.ipynb
```diff
@@ -0,0 +1,467 @@
...
+   "outputs": [],
+   "source": [
+    "import pandas as pd\n",
+    "import numpy as np\n",
+    "import json \n",
+    "\n",
+    "def transform_json_weather_data(filepath):\n",
+    "    f = open(filepath)\n",
+    "    data = json.load(f)\n",
+    "    f.close()\n",
```
nicholasjng, line 167: Please use a context manager:

```python
with open(filepath, "r") as f:
    data = json.load(f)
```

demos/rain_prediction.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "    \n",
+    "df = transform_json_weather_data('./data/weather-2010s.json')\n",
+    "df.head(5)\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Let us now save this data as a csv into the main branch. The `.to_csv` method calls a `put` operation behind the scenes, our commit hook is called and the file committed. You can verify the saving worked in your LakeFS GUI in the browser when looking at the Comits and uncommitted changes tabs of the main branch. "
```
nicholasjng, line 184: This is probably outdated on `main` - the corresponding file event is `FSEvent.FILEUPLOAD` (also, pandas calls `fs.open()` and not `fs.put`).

```suggestion
    "Now we save this data as a CSV file into the main branch. The `DataFrame.to_csv` method calls an `open` operation behind the scenes, our commit hook is called and the file is committed. You can verify the saving worked in the LakeFS UI in your browser by switching to the commits tab of the main branch. "
```

demos/rain_prediction.ipynb
```diff
@@ -0,0 +1,467 @@
...
+   "outputs": [],
+   "source": [
+    "df.to_csv(f'lakefs://{REPO_NAME}/main/weather_2010s.csv')"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We will now do a train test split."
```
nicholasjng, line 200:
```suggestion
    "Now we will do a train-test split."
```

demos/rain_prediction.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "model_data=df.drop('time', axis=1)\n",
+    "\n",
+    "train, test = sklearn.model_selection.train_test_split(model_data)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We save these train and test datasets into a new `training` branch. If, as in this case, the branch does not yet exist, it is implicitly created. You can steer this behaviour via the `create_branch_ok` flag when initializing the `LakeFSFileSystem`. The flag defaults to `True`.  "
```
nicholasjng, line 220:
```suggestion
    "We save these train and test datasets into a new `training` branch. If the branch does not yet exist, as in this case, it is implicitly created by default. You can control this behaviour with the `create_branch_ok` flag when initializing the `LakeFSFileSystem`.  "
```

demos/rain_prediction.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "TRAINING_BRANCH = 'training'\n",
+    "train.to_csv(f'lakefs://{REPO_NAME}/{TRAINING_BRANCH}/train_weather.csv')\n",
+    "test.to_csv(f'lakefs://{REPO_NAME}/{TRAINING_BRANCH}/test_weather.csv')\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The implicit branch creation is a convenient way to create new branches programmatically. However, one drawback is that typos might result in new branch creation. If you want these to throw errors instead, disable the implicit branch creation"
```
nicholasjng, line 238:
```suggestion
    "Implicit branch creation is a convenient way to create new branches programmatically. However, one drawback is that typos in your code might result in new accidental branch creations. If you want to avoid this implicit behavior and raise errors instead, you can disable implicit branch creation by setting `fs.create_branch_ok=False`."
```

demos/rain_prediction.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "test_acc = model.score(x_test, y_test)\n",
+    "\n",
+    "print(f\"Test accuracy: {round(test_acc, 4) * 100 } %\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Until now, we only have used data from the 2010s. Lets get additional 2020s data, transform it and save it to LakeFS. "
```
nicholasjng, line 292:
```suggestion
    "Until now, we only have used data from the 2010s. Let's download additional 2020s data, transform it, and save it to LakeFS. "
```

demos/rain_prediction.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "print(f\"Test accuracy: {round(acc_2020s, 4) * 100 } %\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We have an accuracy similar to the one we had on the 2020s data. Yet, it makes sense to utilize this data for training as well. We will create a concatenated dataframe and perform a new train test split. \n",
+    "\n",
+    "However, this gives rise to the problem, that we now have multiple models which will perfom differently on different datasets. For example, if someone takes the model we are about to train and evaluates it on the data from the 2020s the accuracy will probably be higher. But it will be higher because of data leakage. We are going to use some of the data points in the 2020s data to train. \n",
```
nicholasjng, line 342:
```suggestion
    "However, this means that we now have multiple models which will perform differently on different datasets. For example, if someone takes the model we are about to train and evaluates it on the 2020s data, the accuracy will probably be higher, because of data leakage. We are going to use some of the data points in the 2020s data to train. \n",
```

demos/rain_prediction.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "\n",
+    "train_df.to_csv(f'lakefs://{REPO_NAME}/{TRAINING_BRANCH}/train_weather.csv')\n",
+    "test_df.to_csv(f'lakefs://{REPO_NAME}/{TRAINING_BRANCH}/test_weather.csv')"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This now concatenated the old data, created a new train test split and overwrote the files. This presents problems with respect to strict versioning. When we get the data using only the branch name and filename we get the latest commit. \n",
```
nicholasjng, line 373:
```suggestion
    "This concatenates the old data, creates a new train-test split, and overwrites the files. Of course, this presents problems with respect to strict versioning. When we get the data directly from a branch, we only get the version from the latest commit. \n",
```

demos/rain_prediction.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "\n",
+    "Lets use explicit versioning and get the explicit commit. Therefore go into the LakeFS GUI, select the training branch and choose the \"Commits\" tab. \n",
```
nicholasjng, line 375:
```suggestion
    "Let's use explicit versioning instead and get the actual commit SHA. For that, go into the lakeFS UI, select the training branch, and navigate to the \"Commits\" tab. \n",
```

demos/rain_prediction.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "\n",
+    "You should see the latest two Commits `Add file test_weather.csv` and `Add file train_weather.csv`.\n",
```
nicholasjng, line 377:
```suggestion
    "You should see the latest two commits `Add file test_weather.csv` and `Add file train_weather.csv`.\n",
```

demos/rain_prediction.ipynb
```diff
@@ -0,0 +1,467 @@
...
+   "source": [
+    "test_commit_id  = ''\n",
+    "train_commit_id = ''"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We can now get the specific dataset versions irrespective of subsequent changes to the files on the branches"
```
nicholasjng, line 396:
```suggestion
    "We can now get the specific dataset versions irrespective of subsequent changes to the files on the branches."
```

demos/rain_prediction.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "print(f\"Test accuracy: {round(test_acc, 4) * 100 } %\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We're done. We have a model trained on the new data. \n",
+    "\n",
+    "Now we can save the commit ids as well as the model and accuracy metrics into an experiment tracking tool of our choice. With this, we achieve reproducible experiments and have a clear trail on what data input resulted in which model weights with the resulting hyperparameter results.  \n",
```
nicholasjng, line 435:
```suggestion
    "Now, we can save the commit SHAs as well as the model and accuracy metrics into an experiment tracking tool of our choice. With this, we achieve reproducible experiments and have a clear trail on what input data and hyperparameters resulted in which set of model weights.  \n",
```

demos/rain_prediction.ipynb
```diff
@@ -0,0 +1,467 @@
...
+    "\n",
+    "If in the future we decide to train another model or switch the data up or invest more time in feature engineering, we can always come back to the current state to study the model performance and draw insights that might help us down the road. "
```
nicholasjng, line 437:
```suggestion
    "If in the future we decide to train another model, use different data, or invest more time in feature engineering, we can always come back to the current state to study the model performance and draw insights that might help us down the road. "
```

nicholasjng (commented):
Mostly style suggestions - I would like to take up the precheck issue as a ticket, and register the commit hook to `fileupload` as well to create the pandas commit.

demos/rain_prediction.ipynb
maxmynter: Opened here https://github.com/appliedAI-Initiative/lakefs-spec/issues/92

PR #68, opened 2023-09-12
feat(fs.commit): Expose commit to FS via path

maxmynter:
Expose a commit to the filesystem via the path to LakeFS. [Code](https://github.com/appliedAI-Initiative/lakefs-spec/pull/68/files#diff-f8235da430cb8c0f32975a91dcd263e02044b2621322a7e6c97182b53649ea1eR289).

Adresses parts of https://github.com/appliedAI-Initiative/lakefs-spec/issues/66.

# Other Changes 
- Renamed the previous `commit` function to `_commit` only for internal use
- Test for the new Commit function

### DIFF
--- a/src/lakefs_spec/spec.py
+++ b/src/lakefs_spec/spec.py
@@ -286,9 +286,19 @@ class LakeFSFileSystem(AbstractFileSystem):
         except FileNotFoundError:
             return None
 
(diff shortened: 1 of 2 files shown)


maxmynter:
Closed based on the discussion in the team.  There is a fix that removed the blocker in the example creation. We will address the feedback from https://github.com/appliedAI-Initiative/lakefs-spec/issues/66 exhaustively in a seperate PR.

PR #86, opened 2023-09-24
Implement directory listings caching for `ls`

### LINKED ISSUE #80: Implement `LakeFSFileSystem.ls` caching
The machinery is already in place, now we just need to store results of `ls` calls in the dircache.

- [ ] Pipe storage_options through file system constructor to dircache.
- [ ] Add a check for rpath in the ls cache before API call.
- [ ] Add a store of an `objects_api.list_objects` result to the dircache to the `ls` method.
- [ ] Add test to confirm cache hit.

nicholasjng:
Caches results of `fs.ls` calls in the class' `DirCache` instance.

As an added bonus, if a file is requested by `fs.ls` that has a cache entry for its parent, but is not yet found in the cache, this implementation will fetch those files and append them in the correct place in the existing cache entry. This "hot-reloading" of cache entries is a little more robust than just raising an exception and leaving the user to manually list the objects again.

Adds a test which verifies that the cache is hit on two identical `fs.ls` calls.

Does not assume cache size control or TTL, so this only checks a cache hit in the default configuration.

Closes #80.

### DIFF
--- a/tests/test_ls.py
+++ b/tests/test_ls.py
@@ -1,5 +1,5 @@
 from lakefs_spec import LakeFSFileSystem
-from tests.util import with_counter
+from tests.util import RandomFileFactory, with_counter
 
 
 def test_paginated_ls(fs: LakeFSFileSystem, repository: str) -> None:
@@ -32,3 +32,33 @@ def test_ls_caching(fs: LakeFSFileSystem, repository: str) -> None:
 
     # assert the second `ls` call hits the cache
     assert counter.count("objects_api.list_objects") == 1
+
+
+def test_ls_stale_cache_entry(
+    fs: LakeFSFileSystem,
+    repository: str,
+    random_file_factory: RandomFileFactory,
+    temp_branch: str,
+) -> None:
+    fs.client, counter = with_counter(fs.client)
+
+    random_file = random_file_factory.make()
+
+    resource = f"{repository}/main/data/"
+
+    fs.ls(resource)
+    assert counter.count("objects_api.list_objects") == 1
+    assert len(fs.dircache) == 1
+    assert tuple(fs.dircache.keys()) == ("data",)
+
+    cache_entry = fs.dircache["data"]
+
+    lpath = str(random_file)
+    rpath = f"{repository}/{temp_branch}/data/{random_file.name}"
+
+    fs.put_file(lpath, rpath)
+
+    res = fs.ls(rpath)
+    assert counter.count("objects_api.list_objects") == 2
+    # is the file now added to the cache entry?
+    assert res[0] in cache_entry
--- a/tests/test_ls.py
+++ b/tests/test_ls.py
@@ -66,7 +66,6 @@ def test_ls_stale_cache_entry(
 def test_ls_no_detail(fs: LakeFSFileSystem, repository: str) -> None:
     fs.client, counter = with_counter(fs.client)
 
-    # TODO: This fails without the trailing slash - because of `cls._parent` (?).
     resource = f"{repository}/main/data"
 
     expected = ["data/lakes.source.md"]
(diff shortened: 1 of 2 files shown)


nicholasjng:
So this is now kind of a hard(ish) problem:

TLDR; What should a path like `repo/branch/data` be parsed as? There is no way of knowing whether data is a file or a directory before asking lakeFS.

In the lakeFS sample data,`data` is a single-file directory. But, piping the path `repo/branch/data` populates the `""` entry in the dir cache, which gives the listing for root. So by misinterpreting the `data` prefix as a file, the wrong cache entry is written. This happens because `LakeFSFileSystem._parent("data") == ""`.

Instead, listing `repo/branch/data/` (with trailing slash) makes everything work fine - the trailing slash makes `LakeFSFileSystem._parent` parse the prefix `data/` as `data`, and thus puts the listing into the `data` entry.

HOWEVER, the backend handles the slash and no-trailing slash cases equally - so just from functionality, there should be no difference in whether you put the slash or not.

This means that the implementation should actually _parse_ the parent from the `objects_api.list_objects` result data. instead of just grabbing it blindly from the prefix.

The parse needs to result in the immediate parent for these cases:

* `ls` output is a single file (just call `LakeFSFileSystem._parent` on the name)
* `ls` output is a directory (parse all parents, pick the lowest one - this should already be the prefix (?)), in this case you can probably go with the `.rstrip("/")`ped prefix?

Specifically, in tests we need to make sure that a requested directory prefix in `ls` is NEVER appended/written to an existing entry as in the `data` case above.

tests/test_ls.py
```diff
@@ -13,3 +14,66 @@ def test_paginated_ls(fs: LakeFSFileSystem, repository: str) -> None:
...
+    res = fs.ls(rpath)
+    assert counter.count("objects_api.list_objects") == 2
+    # is the file now added to the cache entry?
+    assert res[0] in cache_entry
+
+
+def test_ls_no_detail(fs: LakeFSFileSystem, repository: str) -> None:
+    fs.client, counter = with_counter(fs.client)
+
+    # TODO: This fails without the trailing slash - because of `cls._parent` (?).
```
nicholasjng, line 69: So this is now kind of a hard(ish) problem:

TLDR; What should a path like `repo/branch/data` be parsed as? There is no way of knowing whether data is a file or a directory before asking lakeFS.

In the lakeFS sample data,`data` is a single-file directory. But, piping the path `repo/branch/data` populates the `""` entry in the dir cache, which gives the listing for root. So by misinterpreting the `data` prefix as a file, the wrong cache entry is written. This happens because `LakeFSFileSystem._parent("data") == ""`.

Instead, listing `repo/branch/data/` (with trailing slash) makes everything work fine - the trailing slash makes `LakeFSFileSystem._parent` parse the prefix `data/` as `data`, and thus puts the listing into the `data` entry.

HOWEVER, the backend handles the slash and no-trailing slash cases equally - so just from functionality, there should be no difference in whether you put the slash or not.

This means that the implementation should actually _parse_ the parent from the `objects_api.list_objects` result data. instead of just grabbing it blindly from the prefix.

The parse needs to result in the immediate parent for these cases:

* `ls` output is a single file (just call `LakeFSFileSystem._parent` on the name)
* `ls` output is a directory (parse all parents, pick the lowest one - this should already be the prefix (?)), in this case you can probably go with the `.rstrip("/")`ped prefix?

Specifically, in tests we need to make sure that a requested directory prefix in `ls` is NEVER appended/written to an existing entry as in the `data` case above.

nicholasjng: Fixed in [7c29669](https://github.com/appliedAI-Initiative/lakefs-spec/pull/86/commits/7c29669fd4b5c105f1496a30e616c18e12af5371) - since the backend returns results sorted by name (and thus by directory nesting starting with the most shallow), we can just put the cache key as the parent of the first, most shallow result.

maxmynter (approved):

PR #88, opened 2023-09-26
Add new `client_helper` functions

maxmynter:
Add `client_helper` functions for convenience to use the lakeFS client. Add tests for these functions.

Closes https://github.com/appliedAI-Initiative/lakefs-spec/issues/82 by implementing
- [x] Revert
- [x] Add tag
- [x] Get Tags and Commits
- [x] Create Tag (and point to commit)

### DIFF
--- a/src/lakefs_spec/client_helpers.py
+++ b/src/lakefs_spec/client_helpers.py
@@ -3,8 +3,8 @@ import sys
 
 from lakefs_client.client import LakeFSClient
 from lakefs_client.model.commit_creation import CommitCreation
-from lakefs_client.model.tag_creation import TagCreation
 from lakefs_client.model.revert_creation import RevertCreation
+from lakefs_client.model.tag_creation import TagCreation
 
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
@@ -28,23 +28,26 @@ def commit(
 
     client.commits_api.commit(repository=repository, branch=branch, commit_creation=commit_creation)
 
-def create_tag(client: LakeFSClient, repository: str, ref:str, tag:str) -> None:
+
+def create_tag(client: LakeFSClient, repository: str, ref: str, tag: str) -> None:
     tag_creation = TagCreation(tag, ref=ref)
     client.tags_api.create_tag(repository=repository, tag_creation=tag_creation)
 
-def get_tags_and_commits(client: LakeFSClient, repository: str) -> None: 
+
+def get_tags(client: LakeFSClient, repository: str) -> dict:
     return client.tags_api.list_tags(repository=repository)
 
-def merge(client: LakeFSClient,
-    repository: str,
-    source_ref: str,
-    target_branch: str) -> None:
 
-    diff = client.refs_api.diff_refs(repository=repository, left_ref=target_branch, right_ref=source_ref)
+def merge(client: LakeFSClient, repository: str, source_ref: str, target_branch: str) -> None:
+    diff = client.refs_api.diff_refs(
+        repository=repository, left_ref=target_branch, right_ref=source_ref
+    )
     if not diff.results:
         logger.warning("No difference between source and target. Aborting merge.")
         return
-    client.refs_api.merge_into_branch(repository=repository, source_ref=source_ref, destination_branch=target_branch) 
+    client.refs_api.merge_into_branch(
+        repository=repository, source_ref=source_ref, destination_branch=target_branch
+    )
 
 
 def revert(client: LakeFSClient, repository: str, branch: str, parent_number: int = 1) -> None:
--- a/src/lakefs_spec/client_helpers.py
+++ b/src/lakefs_spec/client_helpers.py
@@ -3,6 +3,8 @@ import sys
 
 from lakefs_client.client import LakeFSClient
 from lakefs_client.model.commit_creation import CommitCreation
+from lakefs_client.model.tag_creation import TagCreation
+from lakefs_client.model.revert_creation import RevertCreation
 
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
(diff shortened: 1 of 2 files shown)


src/lakefs_spec/client_helpers.py
```diff
@@ -25,3 +27,33 @@ def commit(
...
     client.commits_api.commit(repository=repository, branch=branch, commit_creation=commit_creation)
+
+
+def create_tag(client: LakeFSClient, repository: str, ref: str, tag: str) -> None:
+    tag_creation = TagCreation(tag, ref=ref)
+    client.tags_api.create_tag(repository=repository, tag_creation=tag_creation)
+
+
+def get_tags_and_commits(client: LakeFSClient, repository: str) -> dict:
+    return client.tags_api.list_tags(repository=repository)
```
nicholasjng, line 38: Why `_and_commits`? Does this list the tag commits too?

src/lakefs_spec/client_helpers.py
```diff
@@ -25,3 +27,33 @@ def commit(
...
+
+
+def merge(client: LakeFSClient, repository: str, source_ref: str, target_branch: str) -> None:
+    diff = client.refs_api.diff_refs(
+        repository=repository, left_ref=target_branch, right_ref=source_ref
+    )
+    if not diff.results:
+        logger.warning("No difference between source and target. Aborting merge.")
+        return
```
nicholasjng, line 47: Does this throw like `commit` does if the diff is empty?

src/lakefs_spec/client_helpers.py
```diff
@@ -25,3 +27,33 @@ def commit(
...
+    client.refs_api.merge_into_branch(
+        repository=repository, source_ref=source_ref, destination_branch=target_branch
+    )
+
+
+def revert(
+    client: LakeFSClient, repository: str, branch: str, revert_to_parent_number: int = 1
```
nicholasjng, line 54: Could you rename the argument to `parent_number`, and then write a small docstring saying something like "Revert a branch by ``parent_number`` commits"?

pyproject.toml
```diff
@@ -113,6 +113,7 @@ ignore = [
 exclude_dirs = ["tests"]
 
 [tool.pytest.ini_options]
+pythonpath = [".", "src"]
```
nicholasjng, line 116: Where does this come from? Should not be necessary or?

maxmynter: In my repo I was getting an Error with pytest without it. 

```bash
ImportError while loading conftest '/Path/to/conftest.py'.
tests/conftest.py:16: in <module>
E   ModuleNotFoundError: No module named 'lakefs_spec'
```

maxmynter: I thought it is nicer to configure the path in the `.toml` rather than setting it up manuall w.r.t. possible future contributors.

src/lakefs_spec/client_helpers.py
maxmynter: Yes

src/lakefs_spec/client_helpers.py
maxmynter: Agree. Can rename this. The tags, however, always point to a specific commit. That is why I came up with the naming. But just `get_tags` is probably easier.

pyproject.toml
nicholasjng: How do you invoke `pytest`? From the contributing guide, everything works fine without this change on my machine - I'm therefore inclined to troubleshoot the issue on your PC and leave it as is.

tests/test_client_helpers.py
```diff
@@ -0,0 +1,107 @@
...
+        client=fs.client, repository=repository, branch=temp_branch, message="Commit File Factory"
+    )
+    tag_name = f"Change_{uuid.uuid4()}"
+    client_helpers.create_tag(
+        client=fs.client, repository=repository, ref=temp_branch, tag=tag_name
+    )
+
+    assert any(
+        commit.get("id") == tag_name for commit in fs.client.tags_api.list_tags(repository).results
+    )
```
nicholasjng, line 29: I get why this is necessary - maybe we should add a `temporary_tag` context manager to avoid cluttering the test repo?

pyproject.toml
maxmynter: I just do `pytest` in the root directory. But sure, I'll leave as is.

tests/test_client_helpers.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def test_get_tags(fs: LakeFSFileSystem, repository: str) -> None:
+    response = client_helpers.get_tags(client=fs.client, repository=repository)
+    assert "pagination" in response
+    assert "results" in response
```
nicholasjng, line 35: This test does not make much sense - you are essentially testing `lakefs_client` code.

Maybe this can be integrated into the `test_create_tag` test?

tests/test_client_helpers.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def test_merge_into_branch(
+    random_file_factory: RandomFileFactory, fs: LakeFSFileSystem, repository: str, temp_branch: str
+) -> None:
+    random_file = random_file_factory.make()
+
+    new_branch = "new_branch_" + "".join(random.choices(string.digits, k=8))
+    source_ref = f"{repository}/{new_branch}/{random_file.name}"
```
nicholasjng, line 44: This leaves a dangling branch - you should use the `temporary_branch` context manager fixture and a `with` statement instead.

tests/test_client_helpers.py
```diff
@@ -0,0 +1,107 @@
...
+    )
+    client_helpers.revert(client=fs.client, repository=repository, branch=temp_branch)
+    assert (
+        len(
+            fs.client.refs_api.diff_refs(
+                repository=repository, left_ref="main", right_ref=temp_branch
+            ).results
+        )
+        == 0
+    )
```
nicholasjng, line 107: Technically OK, you think it might be easier to test on number of commits (from n to n+1 back to n)?

nicholasjng (changes requested):

tests/test_client_helpers.py
maxmynter: Yeah, was unsure about this as well. To be fair the `get_tags` function also does not do much more rather than calling the `lakefs_client` so there is not really much more to test. 

As with integration into the `test_create_tag` test, i think it would make more sense to keep the test logic of two separate functions separated.

tests/test_client_helpers.py
maxmynter: AFAIK the commit revert is implemented as an additional commit. So testing for a reduced number of commits does not work. The increase in commits could then also be due to other changes on the branch. Thats why I think checking for the differences on the branch makes most sense here.

nicholasjng: Of course, that was my bad. But still, n->n+1->n+2 should work, no? We are on a temporary branch, and not in a race condition against anyone, so the number of commits should not increase.

(n would have to be set as the number of commits before the one made in the test.)

tests/test_client_helpers.py
nicholasjng: Then I would say let's drop the test for now.

nicholasjng (approved):
Could you rebase, e.g. as one commit for the helpers change and one for the tests? Otherwise this looks great!

tests/test_client_helpers.py
maxmynter: As per discussion, did not move anything into a fixture as to keep the testing logic in the test. However, to ensure an uncluttered repo, I moved the tag creation into a `try finally` block where the `finally` part takes care of cleanup.

PR #90, opened 2023-09-27
Add rev-parse client helper

maxmynter:
Very small PR adding the `rev-parse` client helper. 

Did not include tests as it would essentially be testing `lakeFS client` code.

### DIFF
--- a/src/lakefs_spec/client_helpers.py
+++ b/src/lakefs_spec/client_helpers.py
@@ -2,6 +2,7 @@ import logging
 import sys
 
 from lakefs_client.client import LakeFSClient
+from lakefs_client.exceptions import NotFoundException
 from lakefs_client.model.commit_creation import CommitCreation
 from lakefs_client.model.revert_creation import RevertCreation
 from lakefs_client.model.tag_creation import TagCreation
@@ -63,3 +64,22 @@ def revert(client: LakeFSClient, repository: str, branch: str, parent_number: in
     client.branches_api.revert_branch(
         repository=repository, branch=branch, revert_creation=revert_creation
     )
+
+
+def rev_parse(
+    client: LakeFSClient,
+    repository: str,
+    ref: str,
+    parent: int = 0,
+) -> str:
+    if parent < 0:
+        raise ValueError(f"Parent cannot be negative, got {parent}")
+    try:
+        revisions = client.refs_api.log_commits(
+            repository=repository, ref=ref, limit=2 * (parent + 1)
+        ).results
+    except NotFoundException:
+        raise RuntimeError(
+            "{ref!r} does not match any revision in lakeFS repository {repository!r}"
+        )
+    return revisions[parent].id
--- a/src/lakefs_spec/client_helpers.py
+++ b/src/lakefs_spec/client_helpers.py
@@ -76,10 +76,10 @@ def rev_parse(
         raise ValueError(f"Parent cannot be negative, got {parent}")
     try:
         revisions = client.refs_api.log_commits(
-            repository=repository, ref=ref, limit=2 * (parent + 1)
+            repository=repository, ref=ref, limit=True, amount=2 * (parent + 1)
         ).results
     except NotFoundException:
         raise RuntimeError(
-            "{ref!r} does not match any revision in lakeFS repository {repository!r}"
+            f"{ref!r} does not match any revision in lakeFS repository {repository!r}"
         )
     return revisions[parent].id
--- a/src/lakefs_spec/client_helpers.py
+++ b/src/lakefs_spec/client_helpers.py
@@ -78,8 +78,12 @@ def rev_parse(
         revisions = client.refs_api.log_commits(
             repository=repository, ref=ref, limit=True, amount=2 * (parent + 1)
         ).results
+        if len(revisions) <= parent:
+            raise ValueError(
+                f"cannot fetch revision {ref}~{parent}: {ref} only has {len(revisions)} parents"
+            )
+        return revisions[parent].id
     except NotFoundException:
         raise RuntimeError(
             f"{ref!r} does not match any revision in lakeFS repository {repository!r}"
         )
-    return revisions[parent].id
--- a/src/lakefs_spec/client_helpers.py
+++ b/src/lakefs_spec/client_helpers.py
@@ -79,7 +79,7 @@ def rev_parse(
             repository=repository, ref=ref, limit=True, amount=2 * (parent + 1)
         ).results
         if len(revisions) <= parent:
-            raise ValueError(
+            raise IndexError(
                 f"cannot fetch revision {ref}~{parent}: {ref} only has {len(revisions)} parents"
             )
         return revisions[parent].id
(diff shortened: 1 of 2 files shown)


src/lakefs_spec/client_helpers.py
```diff
@@ -63,3 +63,8 @@ def revert(client: LakeFSClient, repository: str, branch: str, parent_number: in
     client.branches_api.revert_branch(
         repository=repository, branch=branch, revert_creation=revert_creation
     )
+
+
+def rev_parse(client: LakeFSClient, repository: str, ref: str, parent: int = 0) -> str:
+    revisions = client.refs_api.log_commits(repository=repository, ref=ref).results
```
nicholasjng, line 69: Could we use the `after` and `limit` options here for `log_commits`? That optimizes bandwidth and amount of results (we only need at most `parent` results, if we have a merge then it might go to `2 * parent` (?)).

Just out of interest, have you checked what the `prefixes` and `objects` options do?

src/lakefs_spec/client_helpers.py
```diff
@@ -63,3 +63,8 @@ def revert(client: LakeFSClient, repository: str, branch: str, parent_number: in
     client.branches_api.revert_branch(
         repository=repository, branch=branch, revert_creation=revert_creation
     )
+
+
+def rev_parse(client: LakeFSClient, repository: str, ref: str, parent: int = 0) -> str:
```
nicholasjng, line 68: Should we validate client-side and raise a `ValueError` here if `parent` is negative?

src/lakefs_spec/client_helpers.py
```diff
@@ -63,3 +63,8 @@ def revert(client: LakeFSClient, repository: str, branch: str, parent_number: in
...
+    revisions = client.refs_api.log_commits(repository=repository, ref=ref).results
+    return revisions[parent].id
```
nicholasjng, line 70: This is unsafe, and should be guarded by a check like `if not revisions`.

nicholasjng (changes requested):
Nicely done! Had no idea it was that easy, at least without merge commits in the picture.

src/lakefs_spec/client_helpers.py
maxmynter: Good idea

src/lakefs_spec/client_helpers.py
maxmynter: `prefixes` allows to restrict the returned results to the commits that touched a certain directory. `objects` is the same for files.  I've added both to `rev_parse`. 

When specified, one can further the history to the commits which touched a certain directory or file.

src/lakefs_spec/client_helpers.py
maxmynter: Added

src/lakefs_spec/client_helpers.py
maxmynter: I implemented `limit` with `2*(parent +1)`. I don`t think there is a possibility to check if a merge happened without first fetching the log (which would drive the bandwidth saving ad absurdum). 

Did not use `after`. One would need to know the a commit SHA after which one wants to get the results. When you look up the commit sha, you don't need this helper function, right?

src/lakefs_spec/client_helpers.py
```diff
@@ -1,7 +1,9 @@
 import logging
 import sys
+from typing import List
```
nicholasjng, line 3: Please remove, we are developing for `>=3.10` where you can just use `list`.

src/lakefs_spec/client_helpers.py
```diff
@@ -63,3 +65,21 @@ def revert(client: LakeFSClient, repository: str, branch: str, parent_number: in
...
+def rev_parse(
+    client: LakeFSClient,
+    repository: str,
+    ref: str,
+    parent: int = 0,
+    prefixes: List[str] = [],
+    objects: List[str] = [],
+) -> str:
+    if parent < 0:
+        raise ValueError(f"Parent cannot be negative. Got {parent}")
```
nicholasjng, line 79:
```suggestion
        raise ValueError(f"parent cannot be negative, got {parent}")
```

src/lakefs_spec/client_helpers.py
```diff
@@ -63,3 +65,21 @@ def revert(client: LakeFSClient, repository: str, branch: str, parent_number: in
...
+    revisions = client.refs_api.log_commits(
+        repository=repository, ref=ref, limit=2 * (parent + 1), prefixes=prefixes, objects=objects
+    ).results
+    if not revisions:
+        raise ApiException("Could not fetch commit log from lakeFS client.")
```
nicholasjng, line 84: Seems that it just 404s on an unknown ref, so we might just do this:

```suggestion
    try:
        revisions = client.refs_api.log_commits(
            repository=repository, ref=ref, limit=2 * (parent + 1), prefixes=prefixes, objects=objects
        ).results
    except NotFoundException:
        raise RuntimeError("{ref!r} does not match any revision in lakeFS repository {repository!r}")
```
Could also be a `ValueError`.

(You might have to format the code, I typed this blind in the comment field)

src/lakefs_spec/client_helpers.py
```diff
@@ -63,3 +65,21 @@ def revert(client: LakeFSClient, repository: str, branch: str, parent_number: in
...
     )
+
+
+def rev_parse(
+    client: LakeFSClient,
+    repository: str,
+    ref: str,
+    parent: int = 0,
+    prefixes: List[str] = [],
+    objects: List[str] = [],
```
nicholasjng, line 76: 1) I would not support these arguments until they are necessary, and
2) do not use mutable containers (lists, dicts) as default arguments - either use `list()` or make it a `list[str] | None`.

nicholasjng (changes requested):

maxmynter:
Removed the `prefix` and `objects` args. Switched to `try: except` with not found exception. Updated typing. Added tests.

src/lakefs_spec/client_helpers.py
```diff
@@ -63,3 +64,22 @@ def revert(client: LakeFSClient, repository: str, branch: str, parent_number: in
...
+        raise ValueError(f"Parent cannot be negative, got {parent}")
+    try:
+        revisions = client.refs_api.log_commits(
+            repository=repository, ref=ref, limit=True, amount=2 * (parent + 1)
+        ).results
+    except NotFoundException:
+        raise RuntimeError(
+            f"{ref!r} does not match any revision in lakeFS repository {repository!r}"
+        )
+    return revisions[parent].id
```
nicholasjng, line 85: Can you move this return into the `try` block?

Also, from some local exploration it looks like `amount` is the _maximum_ amount of results - so it is possible to produce an `IndexError` here if you request a `parent` of an existing commit that has fewer than `parent` ancestors.

Suggestion: add an `if len(revisions) < parents` check which throws a `ValueError("cannot fetch revision {ref}~{parent}: {ref} only has {len(revisions} parents")`  (or something like that) if the number of ancestors is lower than requested.

nicholasjng (approved):

PR #95, opened 2023-10-05
Allow LakeFS files to decay standard Blockstore files on request [ENG-92]

maxmynter:
Closes https://github.com/appliedAI-Initiative/lakefs-spec/issues/81

## Notable points
- We disable bandit [here](https://github.com/appliedAI-Initiative/lakefs-spec/pull/95/files#diff-f8235da430cb8c0f32975a91dcd263e02044b2621322a7e6c97182b53649ea1eR489). `remote_url` comes from a trusted source (our backend). Additionally, added a test for wrong protocols above.
-  We should still test this with Azure and AWS backends.
- I did not extract the token from the `staging_location`. This proved unnecessary for the handling S3. We should test that that holds with other backends. 
- I was having trouble to identify valuable things to test as most of the code is LakeFS Client code. So if you think the tests are insufficient, just ping me on what else to add. 
-- E.g. did not test upload with `physical address` because its three lines depending solely on lakeFS client APIs ([see here](https://github.com/appliedAI-Initiative/lakefs-spec/pull/95/files#diff-f8235da430cb8c0f32975a91dcd263e02044b2621322a7e6c97182b53649ea1e)).

## Question to reviewer
- Test support pre- and post breaking change of the `get_config` lakeFS API. Unsure what the best practice here is. [See here](https://github.com/appliedAI-Initiative/lakefs-spec/pull/95/files#diff-36cc64783c72d7b449ddf7aa97612ade2b8123f46bb16719b346e8e57613f88dR27)

## Other Changes
- Removed ports from `lakefs-s3-local`. They have no effect when using `network_mode: host`
- Removed `restart: always` from `network_mode: host` to enable shutdown with `docker-compose down`.

## Todos
- [x] Add tests
- [ ] Test with Azure
- [ ] Test with AWS

### DIFF
--- a/hack/lakefs-s3-local.yml
+++ b/hack/lakefs-s3-local.yml
@@ -13,7 +13,9 @@ services:
       - ./config/s3.json:/etc/seaweedfs/s3.json
 
   lakefs:
-    image: treeverse/lakefs:0.110.0
+    image: treeverse/lakefs:0.112.1
+    ports:
+      - 8000:8000
     network_mode: host
     depends_on:
       - seaweedfs
--- a/hack/lakefs-s3-local.yml
+++ b/hack/lakefs-s3-local.yml
@@ -4,21 +4,16 @@ version: "3"
 
 services:
   seaweedfs:
-      container_name: sandbox-s3
-      image: chrislusf/seaweedfs:3.56
-      ports:
-        - 9001:9001
-      network_mode: host
-      command: server -ip.bind 0.0.0.0 -master.volumeSizeLimitMB=1024 -volume.port=9000 -filer.collection=sandbox -s3 -s3.port 9001 -s3.config=/etc/seaweedfs/s3.json -s3.allowEmptyFolder=true -s3.allowDeleteBucketNotEmpty=true
-      volumes:
-          - seaweedfs-data:/data
-          - ./config/s3.json:/etc/seaweedfs/s3.json
-      restart: always
+    container_name: sandbox-s3
+    image: chrislusf/seaweedfs:3.56
+    network_mode: host
+    command: server -ip.bind 0.0.0.0 -master.volumeSizeLimitMB=1024 -volume.port=9000 -filer.collection=sandbox -s3 -s3.port 9001 -s3.config=/etc/seaweedfs/s3.json -s3.allowEmptyFolder=true -s3.allowDeleteBucketNotEmpty=true
+    volumes:
+      - seaweedfs-data:/data
+      - ./config/s3.json:/etc/seaweedfs/s3.json
 
   lakefs:
     image: treeverse/lakefs:0.110.0
-    ports:
-      - 8000:8000
     network_mode: host
     depends_on:
       - seaweedfs
@@ -37,4 +32,4 @@ services:
       LAKEFS_BLOCKSTORE_DEFAULT_NAMESPACE_PREFIX: s3://sandbox/lakefs/
 
 volumes:
-    seaweedfs-data: {}
+  seaweedfs-data: {}
--- a/hack/lakefs-s3-local.yml
+++ b/hack/lakefs-s3-local.yml
@@ -14,8 +14,6 @@ services:
 
   lakefs:
     image: treeverse/lakefs:0.112.1
-    ports:
-      - 8000:8000
     network_mode: host
     depends_on:
       - seaweedfs
--- a/src/lakefs_spec/spec.py
+++ b/src/lakefs_spec/spec.py
@@ -1,18 +1,23 @@
 import hashlib
 import io
 import logging
+import mimetypes
 import operator
 import os
+import urllib.error
+import urllib.request
 from contextlib import contextmanager
 from pathlib import Path
 from typing import Any, Generator
 
+from fsspec import filesystem
 from fsspec.callbacks import NoOpCallback
 from fsspec.spec import AbstractBufferedFile, AbstractFileSystem
 from fsspec.utils import isfilelike, stringify_path
 from lakefs_client import Configuration
 from lakefs_client.client import LakeFSClient
 from lakefs_client.exceptions import ApiException, NotFoundException
+from lakefs_client.model.staging_metadata import StagingMetadata
 from lakefs_client.models import BranchCreation, ObjectCopyCreation, ObjectStatsList
 
 from lakefs_spec.config import LakectlConfig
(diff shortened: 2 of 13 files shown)


src/lakefs_spec/spec.py
```diff
@@ -441,31 +447,95 @@ def _open(
...
+        else:
+            print("OOOKS")
+            print(self.client.config_api.get_config())
+            print(self.client.config_api.get_config().storage_config)
+            blockstore_type = self.client.config_api.get_config().storage_config.blockstore_type
+            print(blockstore_type)
+        if blockstore_type == "local":
+            raise ValueError(
+                "Cannot write to blockstore of type 'local'. Disable use_blockstore or configure remote blockstore."
+            )
```
nicholasjng, line 465: 1. Remove debug prints please,
2. You can use `from lakefs_spec import lakefs_client_version` instead of the split in L454,
3. I would change the errormsg to something shorter, like e.g. `block store writes not implemented for block store type 'local'`.

nicholasjng (commented):
Only reviewed the code for now, will resume with the tests later.

hack/lakefs-s3-local.yml
```diff
@@ -4,21 +4,16 @@ version: "3"
 
 services:
   seaweedfs:
-      container_name: sandbox-s3
```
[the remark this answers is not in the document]
maxmynter: Yes. When we use `network_mode: host` the port forwardings are ignored in the `.yml` as they are exposed directly on the host network (an thus not forwarded).

AdrianoKF:
Regarding your comment on the token in the staging location: from reading the SDK docs I believe this is not meant for authentication but rather used to link the staging location later on:

>  `token` | `str` | opaque staging token to use to link uploaded object

src/lakefs_spec/spec.py
```diff
@@ -441,31 +447,95 @@ def _open(
...
+                    url=remote_url, data=f, headers=headers, method="PUT"
+                )
+                try:
+                    if not remote_url.lower().startswith("http"):
+                        raise ValueError("Wrong protocol for remote connection")
+                    else:
+                        with urllib.request.urlopen(
+                            request
+                        ):  # nosec [B310:blacklist] # We catch faulty protocols above.
+                            logger.info(f"Successfully uploaded: {lpath} ")
```
AdrianoKF, line 490: Maybe add a log statement before the actual upload as well, so the user sees something in case of slow connections

src/lakefs_spec/spec.py
```diff
@@ -441,31 +447,95 @@ def _open(
...
+    ):
+        print("In Blockstore")
+        if tuple(int(v) for v in __lakefs_client_version__.split(".")) < (0, 111, 0):
+            blockstore_type = self.client.config_api.get_storage_config().blockstore_type
+        else:
+            print("OOOKS")
+            print(self.client.config_api.get_config())
+            print(self.client.config_api.get_config().storage_config)
+            blockstore_type = self.client.config_api.get_config().storage_config.blockstore_type
+            print(blockstore_type)
```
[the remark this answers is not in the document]
maxmynter: Ok, I'm blaming the console logs slipping through on the time i committed that yesterday and that the tests were only failing in the CI pipeline and not locally.

src/lakefs_spec/spec.py
maxmynter: Importing throws: `ImportError: cannot import name 'lakefs_client_version' from partially initialized module 'lakefs_spec' (most likely due to a circular import)`

maxmynter: As for the print statements that slipped through. Should we add a test for print statements to the `pre-commit` checks?

src/lakefs_spec/spec.py
```diff
@@ -441,31 +447,95 @@ def _open(
...
+                    print(f"Failed to upload file: {lpath}", e.code, e.reason)
+        else:
+            remote_url = staging_location.physical_address
+            remote = filesystem(blockstore_type, **storage_options)
+            remote.put_file(lpath, remote_url)
+
+        staging_metadata = StagingMetadata(
+            staging=staging_location,
+            checksum=md5_checksum(lpath, blocksize=self.blocksize),
+            size_bytes=1,
```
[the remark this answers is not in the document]
maxmynter: Nope. Added this

src/lakefs_spec/spec.py
```diff
@@ -441,31 +447,95 @@ def _open(
...
+                try:
+                    if not remote_url.lower().startswith("http"):
+                        raise ValueError("Wrong protocol for remote connection")
+                    else:
+                        with urllib.request.urlopen(
+                            request
+                        ):  # nosec [B310:blacklist] # We catch faulty protocols above.
+                            logger.info(f"Successfully uploaded: {lpath} ")
+                except urllib.error.HTTPError as e:
+                    print(f"Failed to upload file: {lpath}", e.code, e.reason)
```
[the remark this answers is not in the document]
maxmynter: I found it easiest to just create an instance of `APIException` from the URLlib error and then call our error translator with it. Agree?

src/lakefs_spec/spec.py
```diff
@@ -70,6 +75,24 @@ def ensure_branch(client: LakeFSClient, repository: str, branch: str, source_bra
         pass
 
 
+def get_client_blockstore_type(client: LakeFSClient) -> str:
+    """
+    Get the blockstore type from the client.
+    Backwards compatible to breaking config_api change in lakeFSClient version 0.110.1.
```
nicholasjng, line 81: Two things:

- The blockstore type is a server/deployment attribute, which the client is used for to query. Please change the name to `get_blockstore_type`.
- I believe the change came in `v0.111.0` (you also have it correct there down below).

src/lakefs_spec/spec.py
```diff
@@ -70,6 +75,24 @@ def ensure_branch(client: LakeFSClient, repository: str, branch: str, source_bra
...
+
+    Args:
+        client (LakeFSClient): The lakefs client.
+
+    Returns:
+        str: The clients blockstore type.
```
nicholasjng, line 87:
```suggestion
        str: The lakeFS server's blockstore type.
```

src/lakefs_spec/spec.py
nicholasjng: Also, maybe move this to the `util` module as well?

src/lakefs_spec/spec.py
```diff
@@ -444,12 +449,68 @@ def _open(
...
+                        ):  # nosec [B310:blacklist] # We catch faulty protocols above.
+                            logger.info(f"Successfully uploaded {lpath}")
+                except urllib.error.HTTPError as e:
+                    urllib_http_error_as_lakefs_api_exception = ApiException(
+                        status=e.code, reason=e.reason
+                    )
+                    translate_lakefs_error(error=urllib_http_error_as_lakefs_api_exception)
+        else:
+            remote_url = staging_location.physical_address
+            remote = filesystem(blockstore_type, **(storage_options or {}))
```
nicholasjng, line 493: Follow-up note: The lakeFS blockstore name is `azure`, the fsspec registry entry is `az`; so we should add a check along the lines of

```python
if blockstore_type == "azure":
    blockstore_type = "az"
```

nicholasjng (approved):
Looks great!

PR #98, opened 2023-10-09
Update Python development dependencies and lakeFS docker image tags

nicholasjng:
Bumps core and dev dependencies, notably `fsspec`, `lakefs-client` and `urllib3`, the latter being a security advisory issued by dependabot on GitHub.

Bump docker compose files to use the latest stable lakeFS tag. This is currently v0.112.1, which carries some changes relevant to the Python userbase. Concretely, it seems that the `lakefs-client` package will eventually give way to the `lakefs-sdk` package.

### DIFF
--- a/src/lakefs_spec/__init__.py
+++ b/src/lakefs_spec/__init__.py
@@ -1,5 +1,7 @@
 from importlib.metadata import PackageNotFoundError, version
 
+from lakefs_client import __version__ as __lakefs_client_version__
+
 from .spec import LakeFSFile, LakeFSFileSystem
 
 try:
@@ -7,3 +9,6 @@ try:
 except PackageNotFoundError:
     # package is not installed
     pass
+
+lakefs_client_version = tuple(int(v) for v in __lakefs_client_version__.split("."))
+del __lakefs_client_version__
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -13,7 +13,7 @@ from lakefs_client import Configuration
 from lakefs_client.client import LakeFSClient
 from lakefs_client.models import BranchCreation, RepositoryCreation
 
-from lakefs_spec import LakeFSFileSystem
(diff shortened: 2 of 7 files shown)


tests/conftest.py
```diff
@@ -70,7 +70,11 @@ def ensurerepo(lakefs_client: LakeFSClient) -> str:
     if _TEST_REPO in reponames:
         logger.info(f"Test repository {_TEST_REPO!r} already exists.")
     else:
-        storage_config = lakefs_client.config_api.get_storage_config()
+        # TODO (n.junge): Remove when the required `lakefs-client` version is at least v0.111.0.
+        if lakefs_client_version < (0, 111, 0):
```
AdrianoKF, line 74: This is a bit brittle, you might want to replace it with [`packaging.version.Version` from the standard library](https://packaging.pypa.io/en/stable/version.html):

```suggestion
        if lakefs_client_version < Version("0.111.0"):
```

src/lakefs_spec/__init__.py
```diff
@@ -1,9 +1,14 @@
...
+
 from .spec import LakeFSFile, LakeFSFileSystem
 
 try:
     __version__ = version("lakefs-spec")
 except PackageNotFoundError:
     # package is not installed
     pass
+
+lakefs_client_version = tuple(int(v) for v in __lakefs_client_version__.split("."))
```
AdrianoKF, line 13: See my other comment:

```suggestion
lakefs_client_version = packaging.version.parse(__lakefs_client_version__)
```

AdrianoKF (changes requested):

AdrianoKF:
Let's keep an eye on `lakefs-sdk`, but for now I'd also stick with `lakefs-client`, until it is officially discontinued.

AdrianoKF (approved):
As discussed in the daily, let's move ahead with the current implementation.

PR #100, opened 2023-10-13
Run hooks even if precheck skips fs action [ENG-100]

maxmynter:
Closes https://github.com/appliedAI-Initiative/lakefs-spec/issues/92

Execute the registered hooks for `FSEvent.PUT_FILE` and `FS_Event.GET_FILE` even if `precheck=True` skips file operations due to no changes.


nicholasjng:
Sad that there's no `goto` in Python...

nicholasjng (approved):

PR #102, opened 2023-10-13
Small fixes re: blockstore puts

nicholasjng:
* Remove mutable dict default from `put_file`
* Move azure->az change to after the blockstore query to improve readability
* Move blockstore type query to the `presign=False` block to avoid the request when using presigned URLs.


src/lakefs_spec/spec.py
```diff
@@ -452,12 +452,6 @@ def _open(
     def put_file_to_blockstore(
         self, lpath, repository, branch, resource, presign=True, storage_options=None
```
nicholasjng, line 453: Do we flip this to `False` to match what we have in `put_file`?

PR #107, opened 2023-10-16
Change lakeFS package to `lakefs-sdk` in place of `lakefs-client`

nicholasjng:
In preparation for the lakeFS v1 launch, change the package to depend on their newly launched `lakefs-sdk` to keep up to date with their upstream changes.

The import change is largely a name replacement only, with the exception of the model imports now happening from the `lakefs_sdk.models` subpackage.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ classifiers = [
     "Typing :: Typed",
 ]
 
-dependencies = ["fsspec>=2023.6.0", "lakefs-client>=0.105.0"]
+dependencies = ["fsspec>=2023.6.0", "lakefs-sdk>=0.113.0.2"]
 
 dynamic = ["version"]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ classifiers = [
     "Typing :: Typed",
 ]
 
-dependencies = ["fsspec>=2023.6.0", "lakefs-sdk>=0.113.0.2"]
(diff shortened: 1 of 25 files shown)


AdrianoKF:
> So I assume the errors mean that I now need to `read()` the file contents on my own in `objects_api.upload_object` instead of handing off the file descriptor?

Yup, seems so. `upload_object` now takes a `content: bytearray` argument: https://github.com/treeverse/lakeFS/blob/master/clients/python/docs/ObjectsApi.md#upload_object

nicholasjng:
This is an upstream issue, tracked in https://github.com/treeverse/lakeFS/issues/6795.

pyproject.toml
```diff
@@ -32,7 +32,7 @@ classifiers = [
     "Typing :: Typed",
 ]
 
-dependencies = ["fsspec>=2023.6.0", "lakefs-client>=0.105.0"]
+dependencies = ["fsspec>=2023.6.0", "lakefs-sdk>=0.113.0.2"]
```
AdrianoKF, line 35: Let's pin this to 1.0 as a lower bound.

nicholasjng: Done in [f7f349f](https://github.com/appliedAI-Initiative/lakefs-spec/pull/107/commits/f7f349fa5dc88ae82f3eb096cea25e062b2b67f4).

AdrianoKF:
The demo notebook still imports from the old `lakefs_client` package, but it seems only used for a type hint. Can you update it, please?

AdrianoKF (approved):

PR #108, opened 2023-10-16
Move ensure_branch out of _warn_on_[...] check

maxmynter:
Move ensure_branch out of _warn_on_fileupload guard such that it's independent on the warning.

### DIFF
--- a/src/lakefs_spec/spec.py
+++ b/src/lakefs_spec/spec.py
@@ -618,8 +618,8 @@ class LakeFSFile(AbstractBufferedFile):
             **kwargs,
         )
 
-        global _warn_on_fileupload
         if mode == "wb":
+            global _warn_on_fileupload
             if _warn_on_fileupload:
                 warnings.warn(
                     f"Calling `{self.__class__.__name__}.open()` in write mode results in unbuffered "
--- a/src/lakefs_spec/spec.py
+++ b/src/lakefs_spec/spec.py
@@ -8,6 +8,7 @@ import operator
 import os
 import urllib.error
 import urllib.request
+import warnings
 from contextlib import contextmanager
 from pathlib import Path
 from typing import Any, Generator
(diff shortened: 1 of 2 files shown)


src/lakefs_spec/spec.py
```diff
@@ -618,13 +619,15 @@ def __init__(
...
         global _warn_on_fileupload
-        if mode == "wb" and _warn_on_fileupload:
-            logger.warning(
-                f"Calling `{self.__class__.__name__}.open()` in write mode results in unbuffered "
-                "file uploads, because the lakeFS Python client does not support multipart "
-                "uploads. Uploading large files unbuffered can have performance implications."
-            )
-            _warn_on_fileupload = False
+        if mode == "wb":
+            if _warn_on_fileupload:
```
nicholasjng, line 623: Let's move the `global _warn_on_fileupload` into this `if` statement, since it is only relevant there.

tests/test_lakefs_file.py
```diff
@@ -37,9 +37,13 @@ def test_lakefs_file_open_write(
 
     rpath = f"{repository}/{temp_branch}/{random_file.name}"
 
-    # try opening the remote file and writing to it
-    with fs.open(rpath, "wb") as fp:
-        fp.write(orig_text)
+    with pytest.warns(
+        UserWarning,
+        match=r"Calling `LakeFSFile\.open\(\)` in write mode results in unbuffered file uploads, because the lakeFS Python client does not support multipart uploads\. Uploading large files unbuffered can have performance implications\.",
```
nicholasjng, line 42:
```suggestion
        match=r"Calling `LakeFSFile\.open\(\)` in write mode results in unbuffered file uploads.*",
```

In general, it is sensible to limit the `match` argument to the shortest possible piece that still conveys the message, that way you save yourself refactoring pains when you rewrite the error messages :)

nicholasjng (changes requested):

nicholasjng (approved):

PR #109, opened 2023-10-16
Update demo `.ipynb` to current version and include `rev_parse` [ENG-99]

maxmynter:
- Updated to `lakefs-spec` commit `8bd2477dbb466f85c714a8659bb66cd92c53cbc7`. We should couple this with a release or, even better, include the `.ipynb` into the CI workflow
- Updated to current 'lakefs-client` version
- Persist the empty data directory such that the `curl` statements work but the data does not pollute the `demos` directory
- Update the postcommit hook to also include the `FILEUPLOAD` event to support and explain pandas write methods.

### DIFF
--- a/demos/rain_prediction.ipynb
+++ b/demos/rain_prediction.ipynb
@@ -137,7 +137,7 @@
    "metadata": {},
    "source": [
     "## Event Hooks\n",
-    "To commit changes programmatically, we can register a hook. This hook needs to have the signature `(client, context) -> None`, where the `client` is the file system's LakeFS client. The context object contains information about the requested resource. "
+    "To commit changes programmatically, we can register a hook. This hook needs to have the signature `(client, context) -> None`, where the `client` is the file system's LakeFS client. The context object contains information about the requested resource.  Within this hook, we can add a the automatic commit. We will register the hook for the `PUT_FILE` event as well as the `FILEUPLOAD` event. Pandas uses the latter with `to_csv` and hence we commit with `to_csv` as well. "
    ]
   },
   {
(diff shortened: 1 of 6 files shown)


demos/rain_prediction.ipynb
```diff
@@ -157,15 +157,18 @@
     "    commit(client, repository=ctx.repository, branch=ctx.ref, message=commit_message)\n",
     "    \n",
     "\n",
-    "#Register the commit hook to be executed after a PUT_FILE event\n",
-    "fs.register_hook(FSEvent.PUT_FILE, commit_on_put)\n"
+    "#Register the commit hook to be executed after the PUT_FILE and FILEUPLOAD events\n",
```
nicholasjng, line 160: Can you add a space between the # and the comment?

demos/rain_prediction.ipynb
```diff
@@ -157,15 +157,18 @@
...
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "However, when you execute the next cell, you will see a message indicating that the upload of the resource has been skipped because the file is uploaded to lakeFS already, and the checksums match. This is useful when we work with large files to reduce the amount of network traffic. Nonetheless, in this specific situation the `PUT` is not executed and neither is our commit hook. \n"
+    "Now uploading the file will create a commit. Since we already uploaded the file, lakeFS will skip the upload as the checksums of the local and remote file match. The hook will be executed regardless.  \n",
+    "\n",
+    "If we want to execute the upload even for an unchanged file, we can do so by passing `precheck=False` to the `PUT` operation."
```
nicholasjng, line 171:
```suggestion
    "If we want to execute the upload even for an unchanged file, we can do so by passing `precheck=False` to the `fs.put()` operation."
```

demos/rain_prediction.ipynb
```diff
@@ -174,23 +177,14 @@
...
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "fs.put('data/weather-2010s.json',  f'{REPO_NAME}/{NEW_BRANCH_NAME}/weather-2010.json', precheck=False)"
+    "However, expecially when you deal with large files the precheck makes sense to avoid repeated but unnecessary transfer of large files. "
```
nicholasjng, line 187:
```suggestion
    "However, especially when dealing with large files, using prechecks makes sense to avoid repeated unnecessary transfers of large files. "
```

demos/requirements.txt
```diff
@@ -12,8 +12,8 @@ jedi==0.19.0
 joblib==1.3.2
 jupyter_client==8.3.1
 jupyter_core==5.3.1
-lakefs-client==0.109.0
-lakefs-spec @ git+https://github.com/appliedAI-Initiative/lakefs-spec.git@e2d188c937310b6053759e7db56804235c64e352
+lakefs-client==0.113.0.1
+lakefs-spec @ git+https://github.com/appliedAI-Initiative/lakefs-spec.git@8bd2477dbb466f85c714a8659bb66cd92c53cbc7
```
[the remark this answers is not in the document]
maxmynter: For now I would need this commit as it is where I merged the fix for the `to_csv` postcommit hooks using pandas. Once we have a release including this, I am all for specifying a stable release or even the latest version once we have the notebooks in the CI pipeline.

PR #110, opened 2023-10-19
Create bare-bones user documentation [ENG-110]

AdrianoKF:
This PR adds a bare-bones documentation setup based on [mkdocs](https://mkdocs.org) and [mkdocs-material](https://squidfunk.github.io/mkdocs-material/):

- Symlinked existing docs (README, contribution guide) to have some content
- Automatically generated API reference using [mkdocstrings](https://mkdocstrings.github.io/), see the [recipe](https://mkdocstrings.github.io/recipes/#automatic-code-reference-pages)
- Execution of Jupyter notebooks with [mknotebooks](https://github.com/greenape/mknotebooks)
- Versioning using [mike](https://github.com/jimporter/mike)
- CI action for publishing to Github pages
- Added a section on documentation to the contribution guide

Built documentation is deployed to GitHub pages for the project at <https://appliedai-initiative.github.io/lakefs-spec/development> (until a `latest` release is produced from a release pipeline, then the `/development` path will become unnecessary). I've requested registration of a custom domain for the project with our IT department already.

## Usage

See also the changes to `CONTRIBUTING.md`.

Install dependencies from `requirements-docs.txt` and run `mkdocs serve` to view docs locally (with instant rebuild), or `mkdocs build` to build the documentation folder under `public/`. See the [mike](https://github.com/jimporter/mike) documentation for details on how versioned docs are built.

## Notes on versioning

The CI workflow for PRs and the `main` branch builds a documentation version with the alias `development`. This means, that parallel PRs and builds for main will race for that alias. Also, old development versions are not cleaned up automatically, which we should/could address in a follow-up PR. (❓ review question: do we want this behavior for PRs? Instead, we could just always deploy to the `develop` version from `main`, and never worry about cleanup)

Versions are determined automatically using the same logic as for package builds, by invoking `setuptools_scm` on the CLI as part of the CI pipeline.

## Dependencies

- `lakefs-client` has been updated to v1.0.0 and made consistent across all lockfiles

## Misc

Besides the functional changes to the CI pipelines for docs builds, repeated code for setting up Python and dependencies has been extracted into a custom GitHub action under `.github/actions/python-deps` to prevent inconsistencies between regular and publishing pipelines in the future.


mkdocs.yml
```diff
@@ -0,0 +1,106 @@
...
+
+# To validate all internal links exist. Does not work in ipynb files
+strict: false
+
+nav:
+  - Introduction: README.md
+  - Use Cases: use-cases.md
+  - User Guide: guides/
+  - API Reference: reference/
+  - CONTRIBUTING.md
```
nicholasjng, line 15: Contribution / Developer Guide maybe?

mkdocs.yml
```diff
@@ -0,0 +1,106 @@
...
+      nav_file: SUMMARY.md
+  - section-index
+  - mkdocstrings:
+      handlers:
+        python:
+          paths: [src]
+          options:
+            docstring_style: numpy
+            docstring_section_style: spacy
+            line_length: 88
```
nicholasjng, line 31: The line length for `black` and `ruff` is 100 in this project, configured in the `pyproject.toml`.

pyproject.toml
```diff
@@ -49,6 +49,19 @@ dev = [
...
+    "mkdocs-gen-files",
+    "mkdocs-literate-nav",
+    "mkdocs-section-index",
+    "mkdocs-material",
+    "mkdocstrings[python]",
+    "mkdocs-git-revision-date-localized-plugin",
+    "mkdocs-include-dir-to-nav",
+    "mknotebooks",
+    "jupyter",
+    "mike@git+https://github.com/jimporter/mike.git",
```
nicholasjng, line 63: Any reason we're not using a stable release (tag etc.)? Does this have a PyPI package?

mkdocs.yml
AdrianoKF: Fixed in 969313a

docs/_scripts/gen_api_ref_pages.py
```diff
@@ -0,0 +1,32 @@
...
+import mkdocs_gen_files
+
+nav = mkdocs_gen_files.Nav()
+
+for path in sorted(Path("src").rglob("*.py")):
+    module_path = path.relative_to("src").with_suffix("")
+    doc_path = path.relative_to("src").with_suffix(".md")
+    full_doc_path = Path("reference", doc_path)
+
+    parts = list(module_path.parts)
```
nicholasjng, line 14: If I understand correctly, you are using `parts` read-only below, so it can stay as a tuple, but your call.

docs/_scripts/gen_api_ref_pages.py
```diff
@@ -0,0 +1,32 @@
...
+
+    if parts[-1] == "__init__":
+        parts = parts[:-1]
+        doc_path = doc_path.with_name("index.md")
+        full_doc_path = full_doc_path.with_name("index.md")
+    elif parts[-1] == "__main__":
+        continue
```
nicholasjng, line 21: I think this might be clearer as an `if` on `module_path.name`?

So let's say:

```python
if module_path.name == "__init__":
        # climb up one level for the module name
        module_path = module_path.parent
        doc_path = doc_path.with_name("index.md")
        full_doc_path = full_doc_path.with_name("index.md")
elif module_path.name == "__main__":
        continue
```

And then you continue below as normal with the now-correctly-truncated module path parts.

pyproject.toml
nicholasjng: Yeah, he does not even have newer git tags available. Then it's fine.

docs/_scripts/gen_api_ref_pages.py
AdrianoKF: This code comes verbatim from the [recipe in the `mkdocstrings` docs](https://mkdocstrings.github.io/recipes/#automatic-code-reference-pages), since it works, I didn't further refine it 😅

docs/_scripts/gen_api_ref_pages.py
AdrianoKF: See above. I've added a link to the original source for the script in 2300f44.

nicholasjng: Eh, I still feel like it's better the other way, but this discussion probably belongs to `mkdocs` then :p

pyproject.toml
```diff
@@ -48,6 +48,21 @@ dev = [
     "build>=0.10.0",
     "pytest-cov>=4.1.0",
     "pyyaml>=6.0.1",
+    "setuptools-scm[toml]>=7.1",
```
nicholasjng, line 51: This is already listed in the build system requirements (L2), so why is it necessary here?

nicholasjng (commented):
I for one am in favor of only keeping the `main@HEAD` documentation on develop (and the stable release tags as well).

pyproject.toml
AdrianoKF: I'd also prefer not to have it here again, but it's needed to derive the current version number from the CLI when building the docs in the CI pipeline (`python -m setuptools_scm`).

I'm open to other suggestions to get it (e.g., from `build`, but that doesn't seem to support doing anything other than building the wheel/sdist, which would be overkill when trying to just get the version number).

AdrianoKF:
> I for one am in favor of only keeping the `main@HEAD` documentation on develop (and the stable release tags as well).

One thing we'd lose in this setup is the ability to examine documentation output on PRs (but one could argue that this can be done locally if needed).

In any case, we should do an `mkdocs build` (or `mike deploy` without `--push`), in order to catch any errors that arise from the docs build (such as failing notebook execution).

I'll prepare a change to the CI pipeline that covers this scenario.

pyproject.toml
nicholasjng: Fair enough, this should be the same package in any reasonable scenario.

(The only reasonable alternative would I guess be a raw `python -c "import lakefs_spec.__version__"` or something, but I can see why this is not as attractive)

AdrianoKF: True, that would be an alternative. However, its outputs differ from `setuptools_scm` in case of a dirty working copy (not sure if it would matter for CI runs in a clean repo):

image

nicholasjng (approved):

PR #120, opened 2023-10-27
Add fsspec callback support for `put_file` and `get_file`

nicholasjng:
Sets file sizes appropriately before uploads/downloads, and does relative updates for downloads.

Unfortunately, since the lakeFS server does not support multipart uploads, we do not get callback support for the `put_file` case; we just add a single 100%-chunk update after upload.

For the case of blockstore writes, we just pass the callback to the resulting file system's `put_file` implementation.

### DIFF
--- a/src/lakefs_spec/spec.py
+++ b/src/lakefs_spec/spec.py
@@ -307,9 +307,6 @@ class LakeFSFileSystem(AbstractFileSystem):
 
         info = self.info(rpath)
 
-        filesize = info["size"]
-        callback.set_size(filesize)
-
         if precheck and Path(lpath).exists():
             local_checksum = md5_checksum(lpath, blocksize=self.blocksize)
             remote_checksum = info.get("checksum")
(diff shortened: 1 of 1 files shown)


nicholasjng:
Testing script (requires the local S3 Docker Compose file and AWS mock credentials set, check the `hack/README.md` for details):

```python
# prerequisite: python -m pip install --upgrade tqdm

from fsspec.callbacks import TqdmCallback


def upload():
    fs = LakeFSFileSystem()

    callback = TqdmCallback()
    fs.put_file(".sandbox/mnist.py", "quickstart/main/mnist.py", use_blockstore=True, precheck=False, callback=callback)
    fs.get_file("quickstart/main/lakes.parquet", ".sandbox/lakes.parquet", precheck=False, callback=callback)

if __name__ == "__main__":
    upload()
```

maxmynter (commented):
Two small comments. 

1 ) On naming: To me, callback seems like it could be a generic callback, but we call specific methods on it. Should we name it something along the lines of `tdqm_callback`.

2) Should we add the callback usage to the docs?

PR #123, opened 2023-10-30
Add type hints to lakeFS file and file system [ENG-123]

nicholasjng:
These type hints are the strictest they can be without excessive errors.

This leads to a few relaxations, though:

* The `outfile` argument was reduced to `typing.Any`, since it is optionally assigned to a str or IO class, which is hard to capture and model correctly.
* Paths are currently limited to `str`s, since they are fed directly to the lakeFS URI regex parser.

Followup PRs could e.g. add support for `PathLike`s as {r,l}paths.

### DIFF
--- a/src/lakefs_spec/spec.py
+++ b/src/lakefs_spec/spec.py
@@ -10,10 +10,10 @@ import urllib.request
 import warnings
 from contextlib import contextmanager
 from pathlib import Path
-from typing import Any, Generator
+from typing import Any, Generator, Literal
 
 from fsspec import filesystem
-from fsspec.callbacks import NoOpCallback
+from fsspec.callbacks import Callback, NoOpCallback
 from fsspec.spec import AbstractBufferedFile, AbstractFileSystem
 from fsspec.utils import isfilelike, stringify_path
 from lakefs_sdk import Configuration
@@ -250,7 +250,7 @@ class LakeFSFileSystem(AbstractFileSystem):
                 self.run_hook(FSEvent.EXISTS, ctx)
                 return exists
 
-    def cp_file(self, path1, path2, **kwargs):
+    def cp_file(self, path1: str, path2: str, **kwargs: Any) -> None:
         if path1 == path2:
             return
 
@@ -278,13 +278,13 @@ class LakeFSFileSystem(AbstractFileSystem):
 
     def get(
         self,
-        rpath,
-        lpath,
-        recursive=False,
-        callback=_DEFAULT_CALLBACK,
-        maxdepth=None,
-        **kwargs,
-    ):
+        rpath: str,
+        lpath: str,
+        recursive: bool = False,
+        callback: Callback = _DEFAULT_CALLBACK,
+        maxdepth: int = None,
+        **kwargs: Any,
+    ) -> None:
         super().get(
             rpath, lpath, recursive=recursive, callback=callback, maxdepth=maxdepth, **kwargs
         )
@@ -292,13 +292,13 @@ class LakeFSFileSystem(AbstractFileSystem):
 
     def get_file(
         self,
-        rpath,
-        lpath,
-        callback=_DEFAULT_CALLBACK,
-        outfile=None,
-        precheck=True,
-        **kwargs,
-    ):
+        rpath: str,
+        lpath: str,
+        callback: Callback = _DEFAULT_CALLBACK,
+        outfile: Any = None,
+        precheck: bool = True,
+        **kwargs: Any,
+    ) -> None:
         repository, ref, resource = parse(rpath)
 
         def run_get_file_hook():
@@ -386,7 +386,7 @@ class LakeFSFileSystem(AbstractFileSystem):
         self.run_hook(FSEvent.INFO, HookContext.new(path))
         return statobj
 
-    def ls(self, path, detail=True, **kwargs):
+    def ls(self, path: str, detail: bool = True, **kwargs: Any) -> list[Any]:
         path = self._strip_protocol(path)
         repository, ref, prefix = parse(path)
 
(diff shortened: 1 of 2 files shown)


AdrianoKF (commented):
https://github.com/appliedAI-Initiative/lakefs-spec/blob/f50c58f23fa5831c23546c1a2489a9d5bd40797b/src/lakefs_spec/config.py#L17

The return could be typed as `Self` (`from typing_annotations import Self`), see [PEP 673](https://peps.python.org/pep-0673/).

(also looking at the same method, the error message for missing `PyYAML` is slightly inaccurate: it says that the config file exists, even when it might not)

AdrianoKF (commented):
https://github.com/appliedAI-Initiative/lakefs-spec/blob/f50c58f23fa5831c23546c1a2489a9d5bd40797b/src/lakefs_spec/errors.py#L16

Remove `# type: ignore` here (doesn't seem to be required in any case)

nicholasjng:
> (also looking at the same method, the error message for missing `PyYAML` is slightly inaccurate: it says that the config file exists, even when it might not)

Hm, good point. We do in fact only call `LakectlConfig.read()` if the `path` input exists, but to make this portable, the existence check could probably be carried out again in the method?

But this raises the question of how to proceed if `path.exists() is False`. Return an empty object and risk silent errors, or raise?

AdrianoKF (commented):
https://github.com/appliedAI-Initiative/lakefs-spec/blob/f50c58f23fa5831c23546c1a2489a9d5bd40797b/src/lakefs_spec/spec.py#L180

This is untyped

src/lakefs_spec/spec.py
```diff
@@ -386,7 +386,7 @@ def info(self, path: str, **kwargs: Any) -> dict[str, Any]:
         self.run_hook(FSEvent.INFO, HookContext.new(path))
         return statobj
 
-    def ls(self, path, detail=True, **kwargs):
+    def ls(self, path: str, detail: bool = True, **kwargs: Any) -> list[Any]:
```
AdrianoKF, line 389: I'd be fine with `list`, since it is equivalent to `list[Any]`

AdrianoKF:
> > (also looking at the same method, the error message for missing `PyYAML` is slightly inaccurate: it says that the config file exists, even when it might not)
> 
> Hm, good point. We do in fact only call `LakectlConfig.read()` if the `path` input exists (see the file system constructor), but to make this portable, the existence check could probably be carried out again in the method?
> 
> But this raises the question of how to proceed if `path.exists() is False`. Return an empty object and risk silent errors, or raise?

For now, we can get away by (a) rewording the error message, and/or (b) reading the file first and then importing `PyYAML` (this would eliminate a race condition between the existence check and trying to read the file, while still keeping the error message relevant).

AdrianoKF:
Strong agree that we should allow passing `str | os.PathLike[str] | Path` in a follow-up issue (see the [internal handling](https://github.com/fsspec/filesystem_spec/blob/096e9a186ffdb39b26b6afdf38c610d092c85f95/fsspec/utils.py#L327) in `fsspec`.

nicholasjng:
> https://github.com/appliedAI-Initiative/lakefs-spec/blob/f50c58f23fa5831c23546c1a2489a9d5bd40797b/src/lakefs_spec/config.py#L17
> 
> The return could be typed as `Self` (`from typing_annotations import Self`), see [PEP 673](https://peps.python.org/pep-0673/).

How do we do this for Python >=3.11, where it's built in? I would prefer spinning off a follow-up for this, since it would also carry an optional dependency on `typing_extensions`.

AdrianoKF:
> > https://github.com/appliedAI-Initiative/lakefs-spec/blob/f50c58f23fa5831c23546c1a2489a9d5bd40797b/src/lakefs_spec/config.py#L17
> > 
> > The return could be typed as `Self` (`from typing_annotations import Self`), see [PEP 673](https://peps.python.org/pep-0673/).
> 
> How do we do this for Python >=3.11, where it's built in? I would prefer spinning off a follow-up for this, since it would also carry an optional dependency on `typing_extensions`.

AFAIK, `typing_extensions` correctly uses the language builtins for versions that support them and only adds backports for older versions, so no changes would be required. However, I'm okay with moving this into a follow-up issue.

nicholasjng:
> AFAIK, `typing_extensions` correctly uses the language builtins for versions that support them (edit: [it does](https://github.com/python/typing_extensions/blob/9de9fd613913faee5db317f827f5eec3755d8a92/src/typing_extensions.py#L1795-L1813)) and only adds backports for older versions, so no changes would be required. However, I'm okay with moving this into a follow-up issue.

So by "no changes", you mean `import typing_extensions` regardless of Python version? That would mean taking up `typing_extensions` as an unconditional dependency. Is this worth it for these few classmethod calls we have?

AdrianoKF:
> So by "no changes", you mean `import typing_extensions` regardless of Python version? That would mean taking up `typing_extensions` as an unconditional dependency. Is this worth it for these few classmethod calls we have?

Exactly. If it's just about `Self`, we can also ignore it for now (as the current typing correctly captures everything - there's a few other edge cases that only `Self` addresses, but we don't have them here).

src/lakefs_spec/spec.py
nicholasjng: Changed in [cc14128](https://github.com/appliedAI-Initiative/lakefs-spec/pull/123/commits/cc141285497fdb326d1aa4e3aae20edbe28ce25e).

nicholasjng:
> https://github.com/appliedAI-Initiative/lakefs-spec/blob/f50c58f23fa5831c23546c1a2489a9d5bd40797b/src/lakefs_spec/errors.py#L16
> 
> Remove `# type: ignore` here (doesn't seem to be required in any case)

Removed in [cc14128](https://github.com/appliedAI-Initiative/lakefs-spec/pull/123/commits/cc141285497fdb326d1aa4e3aae20edbe28ce25e).

src/lakefs_spec/spec.py
AdrianoKF: No, it's not ;)

nicholasjng: Sorry, meant [cc14128](https://github.com/appliedAI-Initiative/lakefs-spec/pull/123/commits/cc141285497fdb326d1aa4e3aae20edbe28ce25e).

AdrianoKF (approved):

PR #124, opened 2023-10-30
Remove stale lakeFS SDK branch statement

nicholasjng:
Since we require `lakefs-sdk>=1.0.0` in the `pyproject.toml`, we can drop the SDK/client version guard.

### DIFF
--- a/src/lakefs_spec/util.py
+++ b/src/lakefs_spec/util.py
@@ -1,7 +1,9 @@
 import re
 
 from lakefs_sdk import __version__ as __lakefs_sdk_version__
-from lakefs_sdk.client import LakeFSClient
+
+lakefs_sdk_version = tuple(int(v) for v in __lakefs_sdk_version__.split("."))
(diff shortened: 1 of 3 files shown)


src/lakefs_spec/util.py
```diff
@@ -1,7 +1,9 @@
 import re
 
 from lakefs_sdk import __version__ as __lakefs_sdk_version__
-from lakefs_sdk.client import LakeFSClient
+
+lakefs_sdk_version = tuple(int(v) for v in __lakefs_sdk_version__.split("."))
```
AdrianoKF, line 5: Is this even still referenced? (my VSCode doesn't find any references)

nicholasjng: No, but I figured it was worth keeping around if we have to do some version guarding in the future. I can also remove it though.

AdrianoKF (approved):

Issue #128, opened 2023-11-02
Add auto-commit mode


AdrianoKF:
For a new user of the library, having to register a hook in order to make a commit after uploading a file can be a bit daunting, since it requires understanding multiple unrelated concepts. There is also a danger of forgetting to create a commit at all, which might lead to unexpected repository states when using the library.

We currently have `client_utils.commit()`, but that involves pulling the client from a filesystem instance (and thus, does not work easily with `LakeFSFile`) and manually keeping track of repo and branch names.

Instead, we could offer a simple `auto_commit=True` option to a filesystem (or a file, as part of `close()`), which would simply create a new commit on the active branch for all modification operations (i.e., `put`, `rm`, `cp_file`). We could even harness the existing hook system for this and automatically register a `commit_file` hook for the relevant `FSEvent`s. (however, this would probably warrant a revisit of the hook system, such that multiple hooks can be registered independently for a given event to enable composability).

## Proposed API changes

- Introduce an `auto_commit: bool` argument in `LakeFSFileSystem`
- Introduce an `auto_commit: bool` argument in `LakeFSFile`

nicholasjng:
Worth remembering that the `AbstractBufferedFile` (and thus also the `LakeFSFile`) take a boolean `autocommit` argument (https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.spec.AbstractBufferedFile), which has a different meaning (controls auto-writing the buffer on final chunk upload) - we should think of how to name them sufficiently differently (really unfortunate naming collision tbh).

maxmynter:
How about `commit_on_push` or `commit_on_write`?

Maciej818:
Not needed due to introducing transaction feature.

PR #129, opened 2023-11-02
Restructure README [ENG-122]

AdrianoKF:
Summary of changes:

- Add feature _Highlights_ to the introduction
- Made Installation section more concise
- Completely replace Usage examples with shorter, working Python code
  for both FS and library access
- Mention Issues and Discussion under Contributing
- Add license information


README.md
```diff
@@ -1,244 +1,90 @@
...
-```shell
-poetry add lakefs-spec
-```
-
-or, for the development version,
-
-```shell
-poetry add git+https://github.com/appliedAI-Initiative/lakefs-spec.git
-```
+If you want `lakefs-spec` to automatically discover and load credentials from an existing `~/.lakectl.yaml` credentials file on your machine, additionally install the `PyYAML` library.
```
janwillemkl, line 19:
```suggestion
If you want `lakefs-spec` to automatically discover and load credentials from an existing `~/.lakectl.yaml` credentials file on your machine, additionally install the `PyYAML` library (`$ pip install pyyaml`).
```

janwillemkl: Maybe remove this line from the README, and add it on a page listing the different installation options?

README.md
```diff
@@ -1,244 +1,90 @@
...
-raw = pd.read_csv('local-path-to.csv')
-raw.to_csv('lakefs://<lakeFS-uri')
-commit('raw data ingestion')
+duckdb.register_filesystem(fsspec.filesystem("lakefs"))
+res = duckdb.read_parquet("lakefs://quickstart/main/lakes.parquet")
+res.show()
 ```
 
-This commit function creates a commit with a unique SHA which you can get from the lakeFS user interface or withthe `lakefs_spec.client_helpers.get_tags` function.
+For a more comprehensive overview of third-party tools integrating with `fsspec` (and therefore by extension `lakefs-spec`), see the [user guide](https://lakefs-spec.org/latest/guides/overview/) in the documentation.
```
janwillemkl, line 79: Good, but hold off on this one until the documentation actually contains the user guide.

README.md
AdrianoKF: I was considering both of your suggestions but opted to put it in the readme because the examples under Usage require the YAML parsing (I left out the `pip install pyyaml`, since users by that point should hopefully know how to install a package, I don't want to repeat everything for both `pip` and `poetry` again 😅).

In general, I am leaning towards adding a dependency on `PyYAML` to the project dependencies to save our users some headaches. WDYT, @nicholasjng?

README.md
```diff
@@ -1,244 +1,90 @@
...
+Please check [discussions](https://github.com/appliedAI-Initiative/lakefs-spec/discussions) or raise an [issue](https://github.com/appliedAI-Initiative/lakefs-spec/issues) on GitHub for any problems you encounter with the library.
 
-Now, with the data and code version identified by a specific commit sha you will always reproduce the same experiment outcomes, e.g. `acc`.
+For information on the general development workflow, see the [contribution guide](CONTRIBUTING.md).
 
-## Developing and contributing to `lakefs-spec`
+## License
 
-We welcome contributions to the project! For information on the general development workflow, head over to the [contribution guide](CONTRIBUTING.md).
+The `lakefs-spec` project is distributed under the [Apache-2 license](https://github.com/appliedAI-Initiative/lakefs-spec/blob/main/LICENSE).
```
janwillemkl, line 90:
```suggestion
The `lakefs-spec` library is distributed under the [Apache-2 license](https://github.com/appliedAI-Initiative/lakefs-spec/blob/main/LICENSE).
```

README.md
nicholasjng: I disagree - has any user reported any headaches? With all due respect, running `pip install --upgrade pyyaml` once in the console (which is even suggested if you fail to load an existing configfile) is possible for every prospective user of this library.

And since we only require it for an opt-in feature (lakectl configs), adding at as an unconditional requirement is not appropriate.

That being said, if we mention it heavily in the user guide, we can either a) change that to use other auth methods or b) put a small disclaimer in front of the examples (preferred).

README.md
```diff
@@ -1,244 +1,90 @@
...
+# Prepare example local data
+local_path = Path("demo.txt")
+local_path.write_text("Hello, lakeFS!")
 
-⚠️ To be able to read settings from a YAML configuration file, `pyyaml` has to be installed, for example using `pip`:
+# Upload to lakeFS and create a commit
+fs = LakeFSFileSystem()  # will auto-discover config from ~/.lakectl.yaml
+repo_path = f"{REPO}/{BRANCH}/{local_path.name}"
+fs.put(str(local_path), repo_path)
+commit(fs.client, REPO, BRANCH, "Add demo data")
```
nicholasjng, line 44: Great stuff! Makes me wonder if this is easier than the hook abstraction? I love this workflow more than the hook registration tbh (we should ask our users!).

Maybe we can explore transactions as a way of naturally creating commits? fsspec supports them, and it fits the natural versioning flow _very_ well in my opinion.

nicholasjng (commented):
I was going to ask if you were putting the deleted content into another document, but most of it seems like cruft that is at best tangentially interesting for users.

Maybe the precheck verification can be documented somewhere else, but other than that, getting rid of all this is a major improvement in my opinion.

README.md
AdrianoKF: It sits orthogonal to the hooks, IMO. Hooks I would see as the lower-level abstraction, in case you need to do something whenever some event occurs (like notifications, etc.), whereas transactions (or auto-commit mode #128) would allow a more convenient high-level way of sequencing the operations. Both have their valid place, but I think we should put more emphasis on the higher-level abstractions in the long run, since that's part of our USP.

nicholasjng: Okay, but are there enough interesting applications of hooks left if we decide to include support for versioning operations on transactions (i.e. on the bulked files)?

The more I think about it, implementing automatic commits, tagging, etc. in transactions and then directing users towards that is a solid way to go. How many people will be left that require event hook support then?

Also, the transaction concept is probably very much in line with the proposed lakeFS plans, and could be swapped out later with the SDK wrapper function if needed.

README.md
```diff
@@ -4,7 +4,15 @@
...
-Our primary goal is to streamline versioned data operations in lakeFS, enabling seamless integration with popular data science tools such as pandas, polars, and DuckDB directly from Python.
+Our primary goal is to streamline versioned data operations in lakeFS, enabling seamless integration with popular data science tools such as Pandas, Polars, and DuckDB directly from Python.
+
+Highlights:
+
+- high-level abstraction over basic lakeFS repository operations
+- seamless integration into the `fsspec` ecosystem
+- zero-config option through config autodiscovery
+- automatic up-/download management to avoid unnecessary transfers for unchanged files
+- extensibility through event hooks
```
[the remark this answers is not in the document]
AdrianoKF: Addressed in 98a82fdd98e45e4bb128edf5618038d0293de096

README.md
AdrianoKF: While I understand your reluctance to include a new mandatory dependency, I'd like to note that the lakectl config feature is not really opt-in, but rather the default behavior of the filesystem, unless the user customizes the constructor parameters.  If we consider it optional, we should change this behavior - if we want this to be the default, it should be as painless as possible for the user, IMO.

In any way, this PR isn't the place to a make this decision, so I would push this discussion into a follow-up issue.

README.md
AdrianoKF: Resolving this thread in favor of the discussions on the other related issues and PRs.

README.md
```diff
@@ -1,244 +1,103 @@
...
+### High-level: Via third-party libraries
 
-# this time, try to read in the default lakectl config, with http://example-host set as host.
-fs = LakeFSFileSystem()
-print(fs.client._api.configuration.host) # <- prints localhost:8000!
-```
-
-The best way to avoid this is to commit to only using either environment variables or `lakectl` configuration files.
-If you do have to mix both methods, you can clear the instance cache like so:
+A variety of widely-used data science tools are building on `fsspec` to access remote storage resources and can thus work with lakeFS data lakes directly through `lakefs-spec`.
```
maxmynter, line 65: Should we add a link to, or list of tools using `fsspec`. Eg. [this one](https://filesystem-spec.readthedocs.io/en/latest/).

Interesting tools using it might be Pandas, DuckDB, dask, kedro, Pyarrow, dvc, Intake.

maxmynter (approved):

README.md
AdrianoKF: Changed in [35ff4a6](https://github.com/appliedAI-Initiative/lakefs-spec/commit/35ff4a6d0ec34917ee03a911c3fe5a67d435a1d3)

README.md
AdrianoKF: Done in [35ff4a6](https://github.com/appliedAI-Initiative/lakefs-spec/commit/35ff4a6d0ec34917ee03a911c3fe5a67d435a1d3)

README.md
AdrianoKF: Resolved through #141

nicholasjng (approved):

PR #133, opened 2023-11-06
Explicitly check for existence in `LakectlConfig.read()`

### LINKED ISSUE #126: Move YAML file existence check into `LakectlConfig`
> > (also looking at the same method, the error message for missing `PyYAML` is slightly inaccurate: it says that the config file exists, even when it might not)
> 
> Hm, good point. We do in fact only call `LakectlConfig.read()` if the `path` input exists (see the file system constructor), but to make this portable, the existence check could probably be carried out again in the method?
> 
> But this raises the question of how to proceed if `path.exists() is False`. Return an empty object and risk silent errors, or raise?

For now, we can get away by (a) rewording the error message, and/or (b) reading the file first and then importing `PyYAML` (this would eliminate a race condition between the existence check and trying to read the file, while still keeping the error message relevant).

I wouldn't silently swallow the error and try to be too smart as it might lead to unexpected behavior.

Also happy to move this into a separate issue.

_Originally posted by @AdrianoKF in https://github.com/appliedAI-Initiative/lakefs-spec/issues/123#issuecomment-1786967941_

nicholasjng:
Based on the class responsibility model, the `LakectlConfig` class should not assume that existence of the input file has already been checked.

However, in the file system case, this is necessary to gracefully decay into an empty config in case the default config does not exist, or cannot be read.

Fixes #126.

### DIFF
--- a/src/lakefs_spec/config.py
+++ b/src/lakefs_spec/config.py
@@ -1,11 +1,8 @@
 from __future__ import annotations
 
-import logging
+import warnings
 from pathlib import Path
 from typing import Any, NamedTuple
-
-logger = logging.getLogger(__name__)
-logger.setLevel(logging.INFO)
 
 
 class LakectlConfig(NamedTuple):
@@ -21,9 +18,10 @@ class LakectlConfig(NamedTuple):
         try:
             import yaml
         except ModuleNotFoundError:
-            logger.warning(
+            warnings.warn(
                 f"Configuration '{path}' cannot be read because `pyyaml` is not installed. "
                 f"To fix, run `python -m pip install --upgrade pyyaml`.",
+                UserWarning,
             )
             return cls()
 
--- a/src/lakefs_spec/config.py
+++ b/src/lakefs_spec/config.py
@@ -15,12 +15,15 @@ class LakectlConfig(NamedTuple):
 
     @classmethod
     def read(cls, path: str | Path) -> "LakectlConfig":
+        if not Path(path).exists():
+            raise FileNotFoundError(path)
+
         try:
             import yaml
         except ModuleNotFoundError:
             logger.warning(
-                f"Configuration '{path}' exists, but cannot be read because the `pyyaml` package "
-                f"is not installed. To fix, run `python -m pip install --upgrade pyyaml`.",
+                f"Configuration '{path}' cannot be read because `pyyaml` is not installed. "
+                f"To fix, run `python -m pip install --upgrade pyyaml`.",
             )
             return cls()
 
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -2,7 +2,7 @@ import sys
 from pathlib import Path
 
 import pytest
-from pytest import LogCaptureFixture, MonkeyPatch
+from pytest import MonkeyPatch
 
 from lakefs_spec.config import LakectlConfig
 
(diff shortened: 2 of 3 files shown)


src/lakefs_spec/config.py
```diff
@@ -15,12 +15,15 @@ class LakectlConfig(NamedTuple):
 
     @classmethod
     def read(cls, path: str | Path) -> "LakectlConfig":
+        if not Path(path).exists():
+            raise FileNotFoundError(path)
```
AdrianoKF, line 19: This is not covered by a unit test. Can you add a small one (not because that code is complex, but out of good habit ;))?

tests/test_config.py
```diff
@@ -0,0 +1,29 @@
...
+def test_lakectl_config_parsing_without_yaml(
+    monkeypatch: MonkeyPatch, caplog: LogCaptureFixture, temporary_lakectl_config: str
+) -> None:
+    # unset YAML module from sys.modules.
+    monkeypatch.setitem(sys.modules, "yaml", None)
+
+    LakectlConfig.read(Path("~/.lakectl.yaml").expanduser())
+
+    # look for the log record, assert that it has the `pyyaml` install instruction.
+    assert len(caplog.records) > 0
```
AdrianoKF, line 28: Looking at this, I wonder, if the [`warnings` module from the standard library](https://docs.python.org/3/library/warnings.html) would be useful for situations like this. I'll play around with it a bit (but not for this PR, which lgtm)

tests/test_config.py
```diff
@@ -0,0 +1,29 @@
...
+        LakectlConfig.read("aaaaaaaaaaaaaaaa.yaml")
+
+
+def test_lakectl_config_parsing_without_yaml(
+    monkeypatch: MonkeyPatch, caplog: LogCaptureFixture, temporary_lakectl_config: str
+) -> None:
+    # unset YAML module from sys.modules.
+    monkeypatch.setitem(sys.modules, "yaml", None)
+
+    LakectlConfig.read(Path("~/.lakectl.yaml").expanduser())
```
AdrianoKF, line 25: Does that not implicitly assume that the config file must exist on the machine to not spuriously fail the test?

tests/test_config.py
nicholasjng: It probably is, I remember correcting a PR like this from a `logger.warning` to a `warnings.warn` to keep a clean stdout / stderr division.

tests/test_config.py
nicholasjng: Setting up the config file (temporarily) is the exact point of the `temporary_lakectl_config` fixture, see `conftest.py`. So yes, it does assume, but rightfully so, since it is created in the fixture setup.

tests/test_config.py
AdrianoKF: There's even [`logging.captureWarnings`](https://docs.python.org/3/library/logging.html#logging.captureWarnings) to forward warnings to the default logging facility.

nicholasjng: huh, TIL. I will change the log to a `warnings.warn`, give me a minute

nicholasjng: Changed to a `UserWarning` in [cca4088](https://github.com/appliedAI-Initiative/lakefs-spec/pull/133/commits/cca4088efdc345b42b9bc9f1cc160a297e680334).

AdrianoKF: > huh, TIL.

So did I 😄 But I vaguely remembered there being a separate warning mechanism in the stdlib, which got me on the path towards `warnings.warn`.

AdrianoKF: > Changed to a `UserWarning` in [cca4088](https://github.com/appliedAI-Initiative/lakefs-spec/pull/133/commits/cca4088efdc345b42b9bc9f1cc160a297e680334).

Awesome, this is much cleaner!

AdrianoKF (approved):

PR #134, opened 2023-11-06
Switch out black for ruff format

nicholasjng:
While it is considered experimental, it appears extremely capable, much much faster than black, and 99.99% compliant with black. Also, none of the listed intended departures from black style seem to apply to us.

Source on known deviations: https://docs.astral.sh/ruff/formatter/black/


AdrianoKF (approved):
I don't have any gripes with black's performance, but since we already use ruff, it makes sense to unify.

Do you think we should add a section on code style to the contribution guide?

PR #135, opened 2023-11-06
Add lakeFS transactions

nicholasjng:
Introduces a subclass of fsspec's builtin transactions called `LakeFSTransaction`, which in addition to its normal use has a number of predefined common versioning operations.

As the transaction is supposed to be used within the context manager block, the generator signature was changed to return the transaction instance, which is a departure from the fsspec instance.

When versioning operations are requested, they are forwarded to their respective client helper function along with their supplied arguments, and the stack of file objects and versioning ops is unwound in the context manager's exit method.

### DIFF
--- a/src/lakefs_spec/client_helpers.py
+++ b/src/lakefs_spec/client_helpers.py
@@ -2,8 +2,9 @@ from __future__ import annotations
 
 import logging
 
+from lakefs_sdk import BranchCreation
 from lakefs_sdk.client import LakeFSClient
-from lakefs_sdk.exceptions import NotFoundException
+from lakefs_sdk.exceptions import ApiException, NotFoundException
 from lakefs_sdk.models import CommitCreation, RevertCreation, TagCreation
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


nicholasjng:
Design plan:

For reference on fsspec's builtin transactions, see https://filesystem-spec.readthedocs.io/en/latest/features.html#transactions. For its base interface definition, see https://github.com/fsspec/filesystem_spec/blob/master/fsspec/transaction.py.

TLDR: I want to add `SetUp` and `Finalize` callbacks to the transaction's signature, to allow a completely programmable code section (context manager) in which to manipulate files in versioned storage.

The callbacks should take at minimum a `fs: LakeFSFileSystem` argument, since that is also passed in the transaction constructor. After that, everything is fair game.

Use cases of the `SetUp` callback are things like creation of new resources (branches, tags, repos). Uses of the `Finalize` callback include creating commits, pushing tags, merging branches, deleting temporary branches, etc.

nicholasjng:
cc @janwillemkl

I'm not sure it will work out the way I thought. If you check out this branch locally, you'll see that the linter complains about the file system's transaction member taking no arguments, so initializing a transaction with these `setup` and `finalize` callbacks from the file system does not seem to be possible.

Another idea is to stick all client helpers onto the transaction, and allow users to call them that way:

```python
fs = LakeFSFileSystem()

with fs.transaction as tx:
    fs.put_file("hello.txt", "quickstart/main/hello.txt")
    fs.put_file("world.txt", "quickstart/main/hello.txt")
    tx.commit("quickstart", "main", "Add my important txt files")  # <- does not need a client because it has access to `fs` as a member.
```

I believe this is also how the lakeFS guys envisioned it. Vey interested to hear your thoughts.

janwillemkl:
I like the idea of putting the helper functions on the transaction! In fact, I like it better than the hook proposal 😄 

One think that can turn tricky is the order at which helper functions may be called. Maybe it makes sense (for now) to limit it to one commit and merge, and they are (by definition) executed after the files are uploaded?

nicholasjng:
> One think that can turn tricky is the order at which helper functions may be called. Maybe it makes sense (for now) to limit it to one commit and merge, and they are (by definition) executed after the files are uploaded?

We could set them up as a shim to the equivalent client helper function, and put them into a FIFO stack (read: list), which is unwound on exit after upload. I will prepare a demo here, will ping you when ready.

nicholasjng:
@janwillemkl Versioning ops are now available as transaction class methods, please take another look.

NB: The `functools.partial` hack in the versioning operation dispatch is an artefact of the client helpers taking only the `client` argument and not the filesystem. I remember your suggestion to make them all take the file system from before, this might come in handy here as a follow up.

(I should at minimum add `self.fs.client` to the `kw` dicts here though.)

maxmynter (approved):

PR #137, opened 2023-11-09
Defer to super class for `get_file`

nicholasjng:
That part of the implementation was left over from a time when the `LakeFSFile` and `fs.open()` were not implemented. As they are now available, we can defer to the superclass for "normal" file gets.


nicholasjng:
I can officially bend spacetime now, getting coverage misses while achieving higher overall coverage percentage AND deleting lines!

PR #139, opened 2023-11-09
Remove hooks functionality

nicholasjng:
The hook concept has been all but superseded by the new transaction API, which enables users to define a versioning workflow more naturally and make fine-grained decisions on when to do versioning, on which files etc.

Since the hooks are now only relevant for lower-level instrumentation, and we currently have no users using them, we remove them again. In a separate PR, we updated the demo notebook to use transactions everywhere, too.

Even though we only have the predefined client helpers on the transaction, which loses flexibility from the hook concept, the workflow using transactions is so superiorly readable and intuitive that most, if not all users should not be hindered.

Advertises transactions explicitly in the documentation as the tool to do versioning flows with.


nicholasjng:
cc @janwillemkl

PR #140, opened 2023-11-10
Improve lakeFS transactions by using deque, placeholders for commit SHAs

nicholasjng:
Similarly to `fsspec`, the transaction now lives in the `lakefs_spec.transaction` submodule.

Introduces the `Placeholder` generic type for a deferred value that is only available after transaction completion.

This is important for values that are not available immediately after execution of a transaction helper, such as a commit SHA, which has to be computed first by the backend, and this happens only on transaction completion.

To be able to still pass these values to other consumers in the transaction, the `ref` argument was changed to also take a `Placeholder` as an argument.

The placeholder is tracked by the transaction's file+op queue, and populated with the actual return value of the executed underlying client helper after the client helper operation completes.

In subsequent operations consuming these placeholder values (with now populated values), the placeholder values are unpacked before the client helper execution, which keeps the client helper interface clean.

These details are hidden away from the user, who just proceeds to use commit SHAs as normal. However, printing these placeholders during the transaction will not show the value for obvious reasons.

----------------------

**Improve client helpers module by allowing commit objects as inputs**

Since the commit object returned by the lakeFS backend contains useful information, we return them from the `commit` and `rev_parse` helpers instead of just their SHA values.

This means that some interfaces have to be relaxed to allow commits, such as the `create_tag` interface (specifically, its `ref` argument).

The tests were adjusted to reflect this new behavior.

### DIFF
--- a/src/lakefs_spec/transaction.py
+++ b/src/lakefs_spec/transaction.py
@@ -1,0 +1,148 @@
+from collections import deque
+from dataclasses import dataclass
+from functools import partial
+from typing import TYPE_CHECKING, Any, Callable, Generic, TypeVar
+
+from fsspec.spec import AbstractBufferedFile
+from fsspec.transaction import Transaction
+from lakefs_sdk.client import LakeFSClient
+from lakefs_sdk.models import Commit
+
+from lakefs_spec.client_helpers import commit, create_tag, ensure_branch, merge, rev_parse, revert
+
+T = TypeVar("T")
+
+if TYPE_CHECKING:
+    from lakefs_spec import LakeFSFileSystem
+
+    VersioningOpTuple = tuple[Callable[[LakeFSFileSystem], None], Any]
+else:
+    VersioningOpTuple = tuple[Callable[["LakeFSFileSystem"], None], Any]
+
+
+@dataclass
+class Placeholder(Generic[T]):
+    value: T | None = None
+
+    def available(self):
+        return self.value is not None
+
+    def set_value(self, value: T) -> None:
+        self.value = value
+
+    def unwrap(self) -> T:
+        if self.value is None:
+            raise RuntimeError("placeholder unfilled")
+        return self.value
+
+
+def unwrap_placeholders(kwargs: dict[str, Any]) -> dict[str, Any]:
+    return {k: v.unwrap() if isinstance(v, Placeholder) else v for k, v in kwargs.items()}
+
+
+class LakeFSTransaction(Transaction):
+    """A lakeFS transaction model capable of versioning operations in between file uploads."""
+
+    def __init__(self, fs: "LakeFSFileSystem"):
+        """
+        Initialize a lakeFS transaction. The base class' `file` stack can also contain
+        versioning operations.
+        """
+        super().__init__(fs=fs)
+        self.fs: "LakeFSFileSystem"
+        self.files: deque[AbstractBufferedFile | VersioningOpTuple] = deque(self.files)
+
+    def __enter__(self):
+        self.start()
+        return self
+
+    def commit(
+        self, repository: str, branch: str, message: str, metadata: dict[str, str] | None = None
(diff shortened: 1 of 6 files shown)


nicholasjng:
cc @janwillemkl

nicholasjng:
Deques and other changes in this PR were adopted upstream by fsspec, so a) the ideas we pursue are sound and b) we can eliminate some of our boilerplate again once we require a sufficiently new version of `fsspec`.

src/lakefs_spec/transaction.py
```diff
@@ -0,0 +1,144 @@
...
+                    result = op(self.fs.client)
+                    # if the transaction member returns a placeholder,
+                    # fill it with the result of the client helper.
+                    if isinstance(retval, Placeholder):
+                        retval.set_value(result)
+
+        self.fs._intrans = False
+
+    def create_branch(self, repository: str, branch: str, source_branch: str) -> str:
+        op = partial(
```
maxmynter, line 106:
```suggestion
    def create_branch(self, repository: str, branch: str, source_branch: str) -> str:
    """Create a new branch from source branch in a repository."""
        op = partial(
```

src/lakefs_spec/transaction.py
```diff
@@ -0,0 +1,144 @@
...
+    def revert(self, repository: str, branch: str, parent_number: int = 1) -> None:
+        """Revert a previous commit on a branch."""
+        op = partial(revert, repository=repository, branch=branch, parent_number=parent_number)
+        self.files.append((op, None))
+        return None
+
+    def rev_parse(
+        self, repository: str, ref: str | Placeholder[Commit], parent: int = 0
+    ) -> Placeholder[Commit]:
+        def rev_parse_op(client: LakeFSClient, **kwargs: Any) -> Commit:
```
maxmynter, line 127:
```suggestion
    ) -> Placeholder[Commit]:
    """Reverse parse the history of the ref."""
        def rev_parse_op(client: LakeFSClient, **kwargs: Any) -> Commit:
```

src/lakefs_spec/transaction.py
```diff
@@ -0,0 +1,144 @@
...
+
+@dataclass
+class Placeholder(Generic[T]):
+    value: T | None = None
+
+    def available(self):
+        return self.value is not None
+
+    def set_value(self, value: T) -> None:
+        self.value = value
```
maxmynter, line 31: Clarifying question: Where does the `set_value` function get called? Is it under the hood of `fs-spec`, or when does the replacement of the placeholders with the values happen?

maxmynter (approved):

src/lakefs_spec/transaction.py
nicholasjng: "Rev" stands for revision, and only looks at that ref(erence) and not its history.

src/lakefs_spec/transaction.py
nicholasjng: That happens here: 

https://github.com/appliedAI-Initiative/lakefs-spec/pull/140/files#diff-6244f5c90acecfad8e9a92a83e64c21a31bfab360bd819eb87b17f04fd63f7c8R100-R101

In words, placeholders are immediate return values of the transaction versioning helpers (i.e. `tx.commit, tx.rev_parse`), and are stored in the `files` queue together with the `partial`-wrapped client helper.

On unwinding the queue in the transaction completion, the return value of the client helper partial (i.e., the actual computed value which we were waiting for) is assigned to the placeholder.

src/lakefs_spec/transaction.py
nicholasjng: Added the docstrings for `create_branch` and `rev_parse` in [f9c7ba6](https://github.com/appliedAI-Initiative/lakefs-spec/pull/140/commits/f9c7ba6e48b3182259339c081203f52a0e1b5325)

PR #154, opened 2023-11-20
Add `exist_ok` flag to all resource-creating client helpers

### LINKED ISSUE #153: Add `exist_ok` attributes to repository, tag, other creatable lakeFS types
The backend returning HTTP 409 errors (resource already exists) breaks idempotence in e.g. the tutorial notebook, which worsens UX for a docs development flow.

An easy fix is to allow their existence (e.g. by an `exist_ok` attribute like in pathlib), and return the object if it already exists.

nicholasjng:
This ensures that usage can be made idempotent, which (for us) has the tangible advantage of allowing for faster docs development iterations.

The heuristic to detect a backend "exists" error is only the status code 409 for now, though we could also do a substring search on the reason. This however would incur a JSON parse.

Changes the tag creation API to return a `Ref` model object.

The `ensure_branch` was renamed to `create_branch` on account of the creation option `exist_ok` that was added.

Fixes #153.

### DIFF
--- a/src/lakefs_spec/client_helpers.py
+++ b/src/lakefs_spec/client_helpers.py
@@ -8,6 +8,7 @@ from lakefs_sdk.exceptions import ApiException, NotFoundException
 from lakefs_sdk.models import (
     Commit,
     CommitCreation,
+    Ref,
     Repository,
     RepositoryCreation,
     RevertCreation,
@@ -31,7 +32,7 @@ def commit(
         logger.warning(f"No changes to commit on branch {branch!r}, aborting commit.")
         return rev_parse(client, repository, branch, parent=0)
 
-    commit_creation = CommitCreation(message=message, metadata=metadata or {})
+    commit_creation = CommitCreation(message=message, metadata=metadata)
 
     new_commit = client.commits_api.commit(
         repository=repository, branch=branch, commit_creation=commit_creation
@@ -39,16 +40,11 @@ def commit(
     return new_commit
 
 
-def create_tag(client: LakeFSClient, repository: str, ref: str | Commit, tag: str) -> None:
-    if isinstance(ref, Commit):
-        ref = ref.id
-    tag_creation = TagCreation(id=tag, ref=ref)
-    client.tags_api.create_tag(repository=repository, tag_creation=tag_creation)
-
-
-def ensure_branch(client: LakeFSClient, repository: str, branch: str, source_branch: str) -> str:
+def create_branch(
+    client: LakeFSClient, repository: str, name: str, source_branch: str, exist_ok: bool = True
+) -> str:
     """
-    Creates a branch named ``branch`` if not already existent.
+    Create a branch in a lakeFS repository.
 
     Parameters
     ----------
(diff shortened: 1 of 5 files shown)


src/lakefs_spec/transaction.py
```diff
@@ -102,15 +102,13 @@ def complete(self, commit: bool = True) -> None:
 
         self.fs._intrans = False
 
-    def create_branch(self, repository: str, branch: str, source_branch: str) -> str:
+    def create_branch(self, repository: str, name: str, source_branch: str) -> str:
```
AdrianoKF, line 105: Any reason not to add the `exist_ok` flag to the transaction as well? Since this is essentially a wrapper around the client helper, I'd expect the same interface here.

AdrianoKF:
Can you add a test case for `client_helpers.create_repository`? The function is currently not covered by any tests ([Codecov](https://app.codecov.io/gh/appliedAI-Initiative/lakefs-spec/pull/154?src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=appliedAI-Initiative#280d80c51f38ac5081037e9ce1b9b57c-R88)

src/lakefs_spec/transaction.py
nicholasjng: Good catch, added in [68908cc](https://github.com/appliedAI-Initiative/lakefs-spec/pull/154/commits/68908cc1cdaf1ced63ed407b83a8b8413676521f).

src/lakefs_spec/client_helpers.py
```diff
@@ -31,59 +32,88 @@ def commit(
...
+        Name of the newly created branch.
     source_branch: str
         Name of the source branch the new branch is created from.
+    exist_ok: bool
+        Ignore errors if the branch already exists.
 
     Returns
     -------
-    The branch name that was given.
+    The newly created branch name.
```
AdrianoKF, line 64: Reword to include the case it already exists

src/lakefs_spec/client_helpers.py
```diff
@@ -31,59 +32,88 @@ def commit(
...
+) -> Repository:
+    """
+    Create a repository with the given name and storage namespace.
 
+    Important: Due to cleanup issues in the lakeFS backend, creating a repository again after prior
+    deletion under the same name and storage namespace will almost certainly not work.
 
-def get_tags(client: LakeFSClient, repository: str) -> dict:
-    return client.tags_api.list_tags(repository=repository)
+    The exist_ok flag only asserts idempotence in case the repository is not deleted between
```
AdrianoKF, line 88:
```suggestion
    The `exist_ok` flag only asserts idempotence in case the repository is not deleted between
```

AdrianoKF (approved):

src/lakefs_spec/client_helpers.py
nicholasjng: Addressed in [643d821](https://github.com/aai-institute/lakefs-spec/pull/154/commits/643d821ad647f440a4dca82841dcbcd65d9c79b7).

src/lakefs_spec/client_helpers.py
nicholasjng: Addressed in [643d821](https://github.com/aai-institute/lakefs-spec/pull/154/commits/643d821ad647f440a4dca82841dcbcd65d9c79b7).

PR #156, opened 2023-11-21
Add file system usage documentation [ENG-148]

nicholasjng:
Covers the most important file system usages and their corresponding APIs, as well as tips and caveats arising from peculiarities around lakeFS SDK and backend.


AdrianoKF:
The added page needs to be included in the `nav` block of `mkdocs.yml` so it shows up in the page navigation.

docs/guides/usage.md
```diff
@@ -0,0 +1,83 @@
...
+```python
+from lakefs_spec import LakeFSFileSystem
+
+fs = LakeFSFileSystem()
+
+# remote path, then local target path.
+fs.get_file("my-repo/my-ref/file.txt", "file.txt")
+```
+
+In case of a directory in lakeFS, use the `fs.get` API together with the `recursive=True` switch:
```
AdrianoKF, line 68:
```suggestion
In the case of a directory in lakeFS, use the `fs.get` API together with the `recursive=True` switch:
```

docs/guides/usage.md
```diff
@@ -0,0 +1,83 @@
...
+
+This guide contains instructions and code snippets on how to use the lakeFS file system.
+
+## Uploading and downloading files
+
+The arguably most important feature of the file system are file transfers.
+
+### File uploads
+
+To upload a file, you can use the `fs.put()` and `fs.put_file` methods:
```
AdrianoKF, line 11:
```suggestion
To upload a file, you can use the `fs.put()` and `fs.put_file()` methods:
```

AdrianoKF: As a user, this paragraph doesn't make the difference between these two methods clear. Can you elaborate a bit?

docs/guides/usage.md
```diff
@@ -0,0 +1,83 @@
...
+```python
+from lakefs_spec import LakeFSFileSystem
+
+fs = LakeFSFileSystem()
+
+# remote path, then local target path.
+fs.put_file("file.txt", "my-repo/my-ref/file.txt")
+```
+
+If you want to upload a whole directory to lakeFS, use the `fs.put` API together with the `recursive=True` switch:
```
AdrianoKF, line 22:
```suggestion
If you want to upload an entire directory to lakeFS, you can use the `fs.put()` API together with the `recursive=True` switch:
```

docs/guides/usage.md
```diff
@@ -0,0 +1,83 @@
...
+fs = LakeFSFileSystem()
+
+# creates the target directory `dir` in the current directory.
+fs.put("dir", "my-repo/my-ref/dir", recursive=True)
+```
+
+!!! info
+
+    This method of file uploading results in two transfers: Once from the client to the lakeFS server, and once from the lakeFS server to the object storage.
+    This can impact performance if very large files are uploaded frequently. To avoid this performance hit, you can also decide to write the file directly to the underlying object storage:
```
AdrianoKF, line 42: Only once can be enough, if the files are large enough :sweat_smile:

```suggestion
    This can impact performance if very large files are uploaded. To avoid this performance hit, you can also decide to write the file directly to the underlying object storage:
```

docs/guides/usage.md
```diff
@@ -0,0 +1,83 @@
...
+
+    ```python
+    fs = LakeFSFileSystem()
+    
+    # remote path, then local target path.
+    fs.put_file("my-repo/my-ref/file.txt", "file.txt", use_blockstore=True)
+    ```
+
+    Direct lakeFS blockstore uploads require the installation of the corresponding file system implementation through `pip`.
```
AdrianoKF, line 51:
```suggestion
    Direct lakeFS blockstore uploads require the installation of the corresponding `fsspec` file system implementation through `pip`.
```

docs/guides/usage.md
```diff
@@ -0,0 +1,83 @@
...
+    For an S3-based lakeFS deployment, install the `s3fs` package. For Google Cloud Storage (GCS), install the `gcsfs` package.
+    For Azure blob storage, install the `adlfs` package.
+
+### File downloads
+
+To download a file, you can use the `fs.get()` or `fs.get_file` methods:
```
AdrianoKF, line 57:
```suggestion
To download a file, you can use the `fs.get()` or `fs.get_file()` methods:
```

AdrianoKF: Same as above, what's the difference between the two?

AdrianoKF (commented):
Nice so far! However, this page should also elaborate on the remainder of the file system operations: `ls`, `info`, `rm` (am I missing anything else?)

docs/guides/usage.md
```diff
@@ -0,0 +1,83 @@
+# How to use the lakeFS file system
+
+This guide contains instructions and code snippets on how to use the lakeFS file system.
+
```
AdrianoKF, line 4: Maybe add a note somewhere in the introduction on the structure of `rpath`s (i.e., `[repo]/[ref]/...`), so the user understands them whenever they occur in a method signature?

docs/guides/usage.md
nicholasjng: Fixed in [8ca0a64](https://github.com/aai-institute/lakefs-spec/pull/156/commits/8ca0a64d1045e89f9db40f7f03c6f693f2f3891a).

docs/guides/usage.md
nicholasjng: Fixed in [8ca0a64](https://github.com/aai-institute/lakefs-spec/pull/156/commits/8ca0a64d1045e89f9db40f7f03c6f693f2f3891a).

docs/guides/usage.md
nicholasjng: Added a sentence on the distinction in [8ca0a64](https://github.com/aai-institute/lakefs-spec/pull/156/commits/8ca0a64d1045e89f9db40f7f03c6f693f2f3891a).

Since the distinction is largely (or only?) the `recursive` switch, we could also decide to advertise the `get()`, `put()`, `copy()`, and `rm()` bulk operations - your call.

nicholasjng:
Rebased on main.

docs/guides/usage.md
```diff
@@ -0,0 +1,158 @@
...
+## The lakeFS URI structure
+
+In the following subsections, we frequently make use of [lakeFS URIs](https://docs.lakefs.io/understand/model.html#lakefs-protocol-uris) in the example code.
+lakeFS URIs are strings identifying resources in a lakeFS deployment through a unique combination of repository name, lakeFS revision/ref name, and file name relative to the repository root.
+
+As an example, a URI like `repo/main/file.txt` addresses the `file.txt` file on the `main` branch in the repository named `repo`.
```
AdrianoKF, line 10: That's actually a bit un-ergonomic API on our part: while you can use the full URL (including the `lakefs://` scheme), the same can't be used as an rpath. Do you think it would make sense to optionally allow this kind of URI as well?

Technically, the form without the scheme isn't an URI at all (the scheme part is mandatory), so we should call it differently.

> `URI = scheme ":" ["//" authority] path ["?" query] ["#" fragment]`

docs/guides/usage.md
```diff
@@ -0,0 +1,158 @@
+# How to use the lakeFS file system
+
+This guide contains instructions and code snippets on how to use the lakeFS file system.
+
+## The lakeFS URI structure
+
+In the following subsections, we frequently make use of [lakeFS URIs](https://docs.lakefs.io/understand/model.html#lakefs-protocol-uris) in the example code.
+lakeFS URIs are strings identifying resources in a lakeFS deployment through a unique combination of repository name, lakeFS revision/ref name, and file name relative to the repository root.
```
AdrianoKF, line 8:
```suggestion
lakeFS URIs identify resources in a lakeFS deployment through a unique path consisting of repository name, lakeFS revision/ref name, and file name relative to the repository root.
```

docs/guides/usage.md
```diff
@@ -0,0 +1,158 @@
...
+#      /c.csv
+#      /...
+
+# downloads `dir` into the current directory.
+fs.get("my-repo/my-ref/dir", "dir", recursive=True)
+```
+
+## Obtaining info on stored objects
+
+To query (also called `stat`) a single object in a lakeFS repository, use the `fs.info()` API:
```
AdrianoKF, line 88:
```suggestion
To query the metadata of a single object in a lakeFS repository (similar to the POSIX `stat` function), use the `fs.info()` API:
```

docs/guides/usage.md
```diff
@@ -0,0 +1,158 @@
...
+
+my_dir_listing = fs.ls("my-repo/my-ref/my-dir/")
+```
+
+This returns a list of Python dictionaries containing information on the objects contained in the requested directory.
+The returned objects have the same fields set as those returned by a normal `fs.info()` call on a file object.
+
+## Deleting objects from a lakeFS branch
+
+To delete objects from a lakeFS branch, use the `fs.rm_file()` or `fs.rm()` APIs. As before, while the former works only for single files, the latter can be used to remove whole directories with the `recursive=True` option.
```
AdrianoKF, line 125:
```suggestion
To delete objects from a lakeFS branch, use the `fs.rm_file()` or `fs.rm()` APIs. As before, while the former works only for single files, the latter can be used to remove entire directories with the `recursive=True` option.
```

docs/guides/usage.md
```diff
@@ -0,0 +1,158 @@
...
+
+fs = LakeFSFileSystem()
+
+fs.rm_file("my-repo/my-branch/my-file.txt")
+
+# removes the whole `my-dir` directory.
+fs.rm("my-repo/my-branch/my-dir/", recursive=True)
+```
+
+When removing files or directories from a branch, those removal operations are staged changes until another commit is created.
```
AdrianoKF, line 138: While this is correct, it also applies to all of the other operations as well.

I'd turn this remark into an admonition towards the beginning of the page (since it might catch the reader off-guard, who might expect these operations to create commits; compare [#128](https://github.com/aai-institute/lakefs-spec/issues/128)).

Maybe that admonition is also a good place to link to the description of the transaction feature (once that page is written), since it addresses the same concern.

docs/guides/usage.md
```diff
@@ -0,0 +1,158 @@
...
+
+fs.rm_file("my-repo/my-branch/my-file.txt")
+
+# removes the whole `my-dir` directory.
+fs.rm("my-repo/my-branch/my-dir/", recursive=True)
+```
+
+When removing files or directories from a branch, those removal operations are staged changes until another commit is created. 
+
+## Copying files between branches
```
AdrianoKF, line 140: This also applies to copying files within a branch, right?

AdrianoKF:
Can you rename the file containing the guide, please? `usage.md` is a bit too generic, IMO. (make sure to change it in `mkdocs.yml` and `index.md` in the user guides directory as well).

docs/guides/usage.md
nicholasjng: I think the majority of APIs also allow full URIs with scheme (those are stripped away in `_strip_protocol`). However, handling is inconsistent as you say, and we should canonicalize this actually. (Maybe extend the parsing regex to work on `lakefs://`-prefixed paths, too?)

docs/guides/usage.md
AdrianoKF: I like the new version, let's keep it!

docs/guides/usage.md
AdrianoKF: That's what I had in mind, I just don't have all the edge cases in mind right now where this might cause problems. Can you maybe check what the other implementations (`s3fs`, ...) allow?

AdrianoKF:
Adding a link to the [upstream docs](https://filesystem-spec.readthedocs.io/en/latest/usage.html#use-a-file-system) might also be sensible.

That page also explicitly highlights `exists()`, maybe we can add a sentence on that as well to our docs.

docs/guides/usage.md
nicholasjng: Changed all occurrences of `whole` to `entire` in [13917c0](https://github.com/aai-institute/lakefs-spec/pull/156/commits/13917c064afc3816ce4cca9787ebb3ad784d3c98).

docs/guides/usage.md
```diff
@@ -0,0 +1,158 @@
...
+
+## The lakeFS URI structure
+
+In the following subsections, we frequently make use of [lakeFS URIs](https://docs.lakefs.io/understand/model.html#lakefs-protocol-uris) in the example code.
+lakeFS URIs are strings identifying resources in a lakeFS deployment through a unique combination of repository name, lakeFS revision/ref name, and file name relative to the repository root.
+
+As an example, a URI like `repo/main/file.txt` addresses the `file.txt` file on the `main` branch in the repository named `repo`.
+
+In some lakeFS file system operations, directories are also allowed as resource names.
+For example, the URI `repo/main/data/` addresses the `data` directory on the `main` branch in the `repo` repository.
```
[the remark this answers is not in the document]
nicholasjng: Done in [13917c0](https://github.com/aai-institute/lakefs-spec/pull/156/commits/13917c064afc3816ce4cca9787ebb3ad784d3c98).

docs/guides/usage.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Uploading and downloading files
+
+The arguably most important feature of the file system are file transfers.
```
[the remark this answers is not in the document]
nicholasjng: Fixed in [13917c0](https://github.com/aai-institute/lakefs-spec/pull/156/commits/13917c064afc3816ce4cca9787ebb3ad784d3c98).

docs/guides/usage.md
nicholasjng: Fixed in [13917c0](https://github.com/aai-institute/lakefs-spec/pull/156/commits/13917c064afc3816ce4cca9787ebb3ad784d3c98).

docs/guides/usage.md
nicholasjng: Also fixed in [13917c0](https://github.com/aai-institute/lakefs-spec/pull/156/commits/13917c064afc3816ce4cca9787ebb3ad784d3c98).

docs/guides/usage.md
nicholasjng: Added a subsection (not admonition) in [13917c0](https://github.com/aai-institute/lakefs-spec/pull/156/commits/13917c064afc3816ce4cca9787ebb3ad784d3c98).

This is arguably more what we want, since it applies to all stateful file operations, it deserves its own section (which also captures the reader's attention, and thus informs them of it).

I couldn't fit the admonition anywhere to my liking, if you have a suggestion, let me know.

docs/guides/usage.md
nicholasjng: Yes, I added info and a code example of a copy to a branch in [13917c0](https://github.com/aai-institute/lakefs-spec/pull/156/commits/13917c064afc3816ce4cca9787ebb3ad784d3c98).

nicholasjng:
> Adding a link to the [upstream docs](https://filesystem-spec.readthedocs.io/en/latest/usage.html#use-a-file-system) might also be sensible.
> 
> That page also explicitly highlights `exists()`, maybe we can add a sentence on that as well to our docs.

The `exists` bit was added in [18d51ef](https://github.com/aai-institute/lakefs-spec/pull/156/commits/18d51ef9fd8378328c9c06141434383d2d58aac1), where should the fsspec docs go?

AdrianoKF:
> > Adding a link to the [upstream docs](https://filesystem-spec.readthedocs.io/en/latest/usage.html#use-a-file-system) might also be sensible.
> > That page also explicitly highlights `exists()`, maybe we can add a sentence on that as well to our docs.
> 
> The `exists` bit was added in [18d51ef](https://github.com/aai-institute/lakefs-spec/pull/156/commits/18d51ef9fd8378328c9c06141434383d2d58aac1), where should the fsspec docs go?

I would suggest to introduce a new second-level heading after the section on committed changes that wraps all the file system operations that follow (and push those one level down). In the intro to that section, we can link to the upstream docs.

docs/guides/filesystem-usage.md
```diff
@@ -0,0 +1,186 @@
...
+As an example, a URI like `repo/main/file.txt` addresses the `file.txt` file on the `main` branch in the repository named `repo`.
+
+In some lakeFS file system operations, directories are also allowed as resource names.
+For example, the URI `repo/main/data/` (note the trailing slash) refers to the `data` directory on the `main` branch in the `repo` repository.
+
+## On staged versus committed changes
+
+When uploading, copying, or removing files or directories from a branch, those removal operations are staged changes until another commit is created.
+`lakefs-spec` does not create these commits automatically, since it separates file operations from versioning operations rigorously.
+If you want to conduct versioning operations, like creating commits, between file transfers, the best way to do so is by using a **filesystem transaction**.
```
AdrianoKF, line 19: Start a new paragraph here for readability.

I would go ahead and create a `TODO` placeholder page in the user guide for the transactions, so we can already link it here.

docs/guides/filesystem-usage.md
```diff
@@ -0,0 +1,186 @@
...
+lakeFS URIs identify resources in a lakeFS deployment through a unique path consisting of repository name, lakeFS revision/ref name, and file name relative to the repository root.
+
+As an example, a URI like `repo/main/file.txt` addresses the `file.txt` file on the `main` branch in the repository named `repo`.
+
+In some lakeFS file system operations, directories are also allowed as resource names.
+For example, the URI `repo/main/data/` (note the trailing slash) refers to the `data` directory on the `main` branch in the `repo` repository.
+
+## On staged versus committed changes
+
+When uploading, copying, or removing files or directories from a branch, those removal operations are staged changes until another commit is created.
```
AdrianoKF, line 17:
```suggestion
When uploading, copying, or removing files or directories from a branch, those operations will result in staged changes in the repository until a commit is created.
```

docs/guides/filesystem-usage.md
nicholasjng: Done in [8cb6e79](https://github.com/aai-institute/lakefs-spec/pull/156/commits/8cb6e79eb899e72e761a99bcc1827c728e808cfb).

docs/guides/usage.md
nicholasjng: s3fs for one parses the bucket and resource names after stripping the protocol, so maybe calling `_strip_protocol` in the `parse` routine is the way to go here.

docs/guides/filesystem-usage.md
```diff
@@ -0,0 +1,191 @@
...
+If you want to upload an entire directory to lakeFS, you can use the `fs.put()` API together with the `recursive=True` switch:
+
+```python
+# structure:
+#   dir/a.txt
+#      /b.yaml
+#      /c.csv
+#      /...
+
+fs.put("dir", "my-repo/my-ref/dir", recursive=True)
```
AdrianoKF, line 55: Having tried this, does not actually work, but rather raises an error:

```
Traceback (most recent call last):
  File "/work/lakefs-spec/.sandbox/demo.py", line 17, in <module>
    fs.put("data", f"{REPO}/{BRANCH}/data/", recursive=True)
  File "/work/lakefs-spec/.direnv/python-3.11/lib/python3.11/site-packages/fsspec/spec.py", line 1069, in put
    self.put_file(lpath, rpath, **kwargs)
  File "/work/lakefs-spec/src/lakefs_spec/spec.py", line 437, in put_file
    local_checksum = md5_checksum(lpath, blocksize=self.blocksize)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/work/lakefs-spec/src/lakefs_spec/util.py", line 11, in md5_checksum
    with open(lpath, "rb") as f:
         ^^^^^^^^^^^^^^^^^
IsADirectoryError: [Errno 21] Is a directory: '/work/lakefs-spec/data'
```

The root cause is that the internal implementation resolves the local path (`fs.expand_path(lpath, recursive=recursive, maxdepth=maxdepth)` in `AbstractFileSystem.put()`), but that includes the directory itself, not just the files contained therein.

The below seems to do the trick (note that the `**` glob runs into the same problem): 


```suggestion
fs.put("dir/*", "my-repo/my-ref/dir", recursive=True)
```

However, this might be worth following up on, because it makes for some surprising behavior when compared to regular globs.

AdrianoKF: I'll change the occurrences in this file to something that works, but let's still follow up on this later on.

docs/guides/filesystem-usage.md
```diff
@@ -0,0 +1,191 @@
...
+
+```python
+# structure:
+#   dir/a.txt
+#      /b.yaml
+#      /c.csv
+#      /...
+
+# downloads the entire `dir` directory (and subdirectories) into the current directory.
+fs.get("my-repo/my-ref/dir", "dir", recursive=True)
```
AdrianoKF, line 97: This does not work at all, but due to other reasons than for `put(..., recursive=True)`. It seems that the implementation of `AbstractFileSystem.expand_paths` might need to be overwritten for our case since it expands the paths without the `{repo}/{branch}/` prefix:

```
rpaths = self.expand_path(rpath, recursive=recursive, maxdepth=maxdepth)

# rpath = 'repo/main/data/'
# rpaths = ['data/weather-2010.json', 'data/weather-2010s.json', 'data/weather-2020.json', 'data/weather-2020s.json', 'repo/main/data/']
```

This leads to errors down the road when the code enters `get_file` for these rpaths, since they obviously don't `parse()` correctly.

AdrianoKF:
I've merged the fixes for #162, #163, and #164, which were affecting the demo code snippets. Could you check them again if you have time, @maxmynter and/or @janwillemkl?

docs/guides/filesystem-usage.md
```diff
@@ -0,0 +1,191 @@
...
+
+As an example, a URI like `repo/main/file.txt` addresses the `file.txt` file on the `main` branch in the repository named `repo`.
+
+In some lakeFS file system operations, directories are also allowed as resource names.
+For example, the URI `repo/main/data/` (note the trailing slash) refers to the `data` directory on the `main` branch in the `repo` repository.
+
+## On staged versus committed changes
+
+When uploading, copying, or removing files or directories from a branch, those removal operations will result in staged changes in the repository until a commit is created.
+`lakefs-spec` does not create these commits automatically, since it separates file operations from versioning operations rigorously.
```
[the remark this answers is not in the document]
AdrianoKF: I would say it goes a bit beyond the scope of this page

docs/guides/filesystem-usage.md
AdrianoKF: Addressed in 7e09b41c9f7d3ac7a0b2582832432358ebf4ff42, which makes changing the example unnecessary.

docs/guides/filesystem-usage.md
AdrianoKF: Addressed in 7e09b41c9f7d3ac7a0b2582832432358ebf4ff42 and through #166.

docs/guides/filesystem-usage.md
```diff
@@ -0,0 +1,191 @@
...
+
+This function returns `True` if the file exists on that revision, and `False` if it does not. Errors (e.g. permission errors) will be raised, since in that case, object existence cannot be decided.
+
+!!! Warning
+    
+    `fs.exists()` only works on file objects, and will return `False` if called on directories.
+
+### Obtaining info on stored objects
+
+To query the metadata of a single object in a lakeFS repository (similar to the POSIX `stat` function), use the `fs.info()` API:
```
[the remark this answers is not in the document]
AdrianoKF: Ha, this was probably changed by accident, Nico and I had agreed to leave out the POSIX reference.

```suggestion
To query the metadata of a single object in a lakeFS repository, use the `fs.info()` API:
```

AdrianoKF (approved):

PR #158, opened 2023-11-21
Change certain aspects of the data science tutorial notebook

nicholasjng:
Improves a few aspects of the tutorial:

* The weather API data is now saved to a temporary file, no more directories are created. The tempfiles are closed immediately after their last uses in the upload transactions.
* The separate `put` operation section was cut, because the showcase of a commit abort is unnecessary. Instead, the focus is shifted to the transactional upload of the data frames.
* Instead of shape comparisons, `pandas.DataFrame.equal` is used to establish equality of local vs. remote downloaded data. This actually caught a dtype mismatch between two datasets.
* Through style changes and regroups, a number of text passages were slimmed down or cut entirely.

### DIFF
--- a/docs/tutorials/demo_data_science_project.ipynb
+++ b/docs/tutorials/demo_data_science_project.ipynb
@@ -9,40 +9,26 @@
    "source": [
     "# Introduction\n",
     "\n",
-    "In this notebook, we will complete a small end-to-end data science tutorial that employs `lakefs-spec` for data versioning. We will use versioned weather data to train a random forest classifier to predict whether a given day from now is a raining day given the current weather.\n",
+    "In this notebook, we will complete a small end-to-end data science tutorial that employs `lakefs-spec` for data versioning. We will use versioned weather data to train a decision tree classifier to predict whether it is raining tomorrow given the current weather.\n",
     "\n",
     "We will do the following:\n",
     "\n",
     "* Environment setup\n",
     "* LakeFS setup\n",
+    "* Authenticate with the lakeFS client\n",
     "* Data ingestion via transactions\n",
     "* Model training\n",
     "* Updating data and retraining a model\n",
     "* Accessing data versions and reproducing experiments\n",
     "* Using tags for semantic versioning\n",
     "\n",
-    "To execute the code in this tutorial as a Jupyter notebook, download this `.ipynb` file to a convenient location on your machine. You can also clone the whole `lakefs-spec` repository. During the execution of this tutorial, in the same directory, a folder `data` will be created. We will also download a file `.lakectl.yaml` to the current user's home directory.\n",
+    "To execute the code in this tutorial as a Jupyter notebook, download the `.ipynb` file from the `lakefs-spec` repository. \n",
     "\n",
     "This tutorial assumes that you have installed `lakefs-spec` in a virtual environment, and that you have followed the [quickstart guide](../quickstart.md) to set up a local lakeFS instance.\n",
     "\n",
     "## Environment setup\n",
     "\n",
-    "Install the libraries necessary for this notebook on the environment you have just created: "
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "id": "2a1b38a3-dd43-424e-be1a-9c731bdcf4f5",
-   "metadata": {
-    "cell_marker": "\"\"\"",
-    "tags": [
-     "remove_output"
-    ]
-   },
-   "outputs": [],
-   "source": [
-    "%pip install lakefs-spec numpy pandas scikit-learn"
+    "Install the necessary libraries for this notebook on the environment you have just created: `pip install lakefs-spec numpy pandas scikit-learn`."
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


docs/tutorials/demo_data_science_project.ipynb
```diff
@@ -9,25 +9,23 @@
...
     "\n",
     "To execute the code in this tutorial as a Jupyter notebook, download this `.ipynb` file to a convenient location on your machine. You can also clone the whole `lakefs-spec` repository. During the execution of this tutorial, in the same directory, a folder `data` will be created. We will also download a file `.lakectl.yaml` to the current user's home directory.\n",
     "\n",
     "This tutorial assumes that you have installed `lakefs-spec` in a virtual environment, and that you have followed the [quickstart guide](../quickstart.md) to set up a local lakeFS instance.\n",
     "\n",
-    "# Environment Setup\n",
+    "## Environment setup\n",
     "\n",
     "Install the libraries necessary for this notebook on the environment you have just created: "
    ]
```
Maciej818, line 31: I propose to change:  "%pip install lakefs-spec numpy pandas scikit-learn"
To: "%pip install notebook numpy pandas scikit-learn"
Argumentation: lakefs-spec already installed in the quickstart tutorial; assuming a user created vanilla fresh venv, it requires notebook installation.

docs/tutorials/demo_data_science_project.ipynb
```diff
@@ -9,25 +9,23 @@
...
+    "In this notebook, we will complete a small end-to-end data science tutorial that employs `lakefs-spec` for data versioning. We will use versioned weather data to train a random forest classifier to predict whether a given day from now is a raining day given the current weather.\n",
     "\n",
     "We will do the following:\n",
     "\n",
     "* Environment setup\n",
-    "* lakeFS setup\n",
-    "* Data ingestion\n",
-    "    * Transactions\n",
-    "    * PUT a file\n",
+    "* LakeFS setup\n",
```
Maciej818, line 17: I propose to skip this section (not to repeat quickstart doc)

docs/tutorials/demo_data_science_project.ipynb
```diff
@@ -184,8 +155,8 @@
    },
    "outputs": [],
    "source": [
-    "destination = \"data/weather-2010.json\"\n",
-    "urllib.request.urlretrieve(\"https://archive-api.open-meteo.com/v1/archive?latitude=52.52&longitude=13.41&start_date=2010-01-01&end_date=2010-12-31&hourly=temperature_2m,relativehumidity_2m,rain,pressure_msl,surface_pressure,cloudcover,cloudcover_low,cloudcover_mid,cloudcover_high,windspeed_10m,windspeed_100m,winddirection_10m,winddirection_100m\", destination)"
+    "outfile = tempfile.NamedTemporaryFile()\n",
+    "urllib.request.urlretrieve(\"https://archive-api.open-meteo.com/v1/archive?latitude=52.52&longitude=13.41&start_date=2010-01-01&end_date=2010-12-31&hourly=temperature_2m,relativehumidity_2m,rain,pressure_msl,surface_pressure,cloudcover,cloudcover_low,cloudcover_mid,cloudcover_high,windspeed_10m,windspeed_100m,winddirection_10m,winddirection_100m\", outfile.name)"
```
Maciej818, line 159: After running this I got Permission error (not sure whether it is specific issue on my end):
---------------------------------------------------------------------------
PermissionError                           Traceback (most recent call last)
Cell In[8], line 2
      1 outfile = tempfile.NamedTemporaryFile()
----> 2 urllib.request.urlretrieve("https://archive-api.open-meteo.com/v1/archive?latitude=52.52&longitude=13.41&start_date=2010-01-01&end_date=2010-12-31&hourly=temperature_2m,relativehumidity_2m,rain,pressure_msl,surface_pressure,cloudcover,cloudcover_low,cloudcover_mid,cloudcover_high,windspeed_10m,windspeed_100m,winddirection_10m,winddirection_100m", outfile.name)

File ~\AppData\Local\Programs\Python\Python311\Lib\urllib\request.py:251, in urlretrieve(url, filename, reporthook, data)
    249 # Handle temporary file setup.
    250 if filename:
--> 251     tfp = open(filename, 'wb')
    252 else:
    253     tfp = tempfile.NamedTemporaryFile(delete=False)

PermissionError: [Errno 13] Permission denied: 'C:\\Users\\MACIEJ~1\\AppData\\Local\\Temp\\tmp4rroi7te'

nicholasjng: This is certainly Windows-specific, since it passes in CI. Some aspects of the `tempfile` module differ on Windows, so I imagine this is one of them (or Windows gives fewer default rights to the Temp location).

docs/tutorials/demo_data_science_project.ipynb
nicholasjng: So I'm wondering at what stage we assume the user to check this notebook? If the user does not look at the quickstart initially, he is left without any instructions. I think it's worth including some setup in both - what amount that should be can be discussed.

docs/tutorials/demo_data_science_project.ipynb
nicholasjng: This cell is a Jupyter magic instruction, so it will not run if `notebook` is not already installed.

docs/tutorials/demo_data_science_project.ipynb
```diff
@@ -9,25 +9,23 @@
...
     "* Model training\n",
     "* Updating data and retraining a model\n",
-    "* Accessing data versions and Reproducing Experiments\n",
-    "* Using a tag instead of a commit SHA for semantic versioning\n",
+    "* Accessing data versions and reproducing experiments\n",
+    "* Using tags for semantic versioning\n",
     "\n",
     "To execute the code in this tutorial as a Jupyter notebook, download this `.ipynb` file to a convenient location on your machine. You can also clone the whole `lakefs-spec` repository. During the execution of this tutorial, in the same directory, a folder `data` will be created. We will also download a file `.lakectl.yaml` to the current user's home directory.\n",
     "\n",
     "This tutorial assumes that you have installed `lakefs-spec` in a virtual environment, and that you have followed the [quickstart guide](../quickstart.md) to set up a local lakeFS instance.\n",
```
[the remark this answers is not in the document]
nicholasjng: Hmm, and what if the user needs to authenticate first and / or leave the email? Then "ensure you can see the list of repositories" is not a great instruction, since you do not know what to do if you're left without guidance.

Is the YAML file being `urlretrieve`'d a problem? We assume a local quickstart instance, for which this should be fine.

docs/tutorials/demo_data_science_project.ipynb
Maciej818: This worked for me on Windows:
```
# Save fetched json as a temporary file
with tempfile.NamedTemporaryFile(delete=False, mode='w+', suffix='.json') as temp_file:
    temp_file.write(response.text)
    temp_file.seek(0)  # Move the file pointer to the beginning
```

docs/tutorials/demo_data_science_project.ipynb
maxmynter: Up until now, we have always had the imperative that the Demo notebook should work as a standalone unit. If we want to continue with this, we should keep the section. 
I prefer this approach as I, as a user, would expect a demo to be self-contained. 

If we decide against self-containment, we could significantly trim the set-up explanation as well.

maxmynter (approved):

docs/tutorials/demo_data_science_project.ipynb
maxmynter: Currently, this cell execution breaks the notebook as it pulls the latest release of `lakefs-spec` from PyPI. In this release, we do not have e.g. the `create_repository` function. 

This particular problem should be fixed with a new PyPI release. But in general, I would prefer the installation to point to the current version on `main` as to not run into a problem due to the divergence of the GitHub Repo and PyPI again.

docs/tutorials/demo_data_science_project.ipynb
```diff
@@ -9,42 +9,67 @@
...
     "\n",
     "We will do the following:\n",
     "\n",
     "* Environment setup\n",
-    "* lakeFS setup\n",
-    "* Data ingestion\n",
-    "    * Transactions\n",
-    "    * PUT a file\n",
+    "* LakeFS setup\n",
+    "* Authenticate with the lakeFS client\n",
```
nicholasjng, line 18:
```suggestion
    "* Authenticating with the lakeFS server\n",
```

since authentication is done *with* the authority that's verifying you, which is the lakeFS server in this case.

docs/tutorials/demo_data_science_project.ipynb
```diff
@@ -9,42 +9,67 @@
...
+   "source": [
+    "%pip install numpy pandas scikit-learn"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
    "source": [
-    "%pip install lakefs-spec numpy pandas scikit-learn"
+    "Also install the appropriate lakeFS-spec version. That is either the latest release from PyPI via `pip install lakefs-spec`, or the development version from GitHub via `pip install git+https://github.com/aai-institute/lakefs-spec.git`."
```
nicholasjng, line 72: a) Please strip cell outputs before checking in the notebook, and b) isn't the previous instruction preferable here? It works in the case of both a CI install of a checkout, and a normal PyPI install.

In this current form, the notebook will not install lakefs-spec anymore if the user skips through the comments, which is not ideal.

PR #160, opened 2023-11-23
Change `fs.info()` logic to recognize non-slash-terminated directories

nicholasjng:
Previously, our heuristic for calling `objects_api.stat_object` was the resource terminating in a slash.

However, we put this into an `if-else` branch, meaning that `objects_api.list_objects` would never be called on a directory resource that was not slash-terminated. This leads to erroneous "not found" errors on existing directories.

The heuristic now changes to the strategy "call `stat_object` first, return if successful, otherwise call `list_objects`". This ensures that on an unsuccessful `stat_objects` call on a directory, the correct `list_objects` call follows directly afterwards.

Also adds a regression test for the behavior.

------

Repro on main (currently results in "404 not found" as mentioned above, requires a repo "lakefs-spec-tests" with sample data):

```python

def main():
    """
    Regression test to check that calling `fs.info()` on existing non-slash-terminated
    directories yields the same results as if terminated with a slash.
    """
    fs = LakeFSFileSystem()
    resource = "lakefs-spec-tests/main/data/"
    res = fs.info(resource)

    assert res["type"] == "directory"

    non_slash_resource = resource.rstrip("/")
    assert fs.info(non_slash_resource) == res


if __name__ == "__main__":
    main()
```

I think this is worth it since we all hate websites that behave differently on trailing-slash vs. no trailing slash requests.

### DIFF
--- a/tests/test_ls.py
+++ b/tests/test_ls.py
@@ -75,7 +75,7 @@ def test_ls_stale_cache_entry(
     lpath = str(random_file)
     rpath = f"{repository}/{temp_branch}/data/{random_file.name}"
 
-    fs.put_file(lpath, rpath)
+    fs.put_file(lpath, rpath, precheck=False)
 
     res = fs.ls(rpath)
     assert counter.count("objects_api.list_objects") == 2
--- a/src/lakefs_spec/spec.py
+++ b/src/lakefs_spec/spec.py
@@ -251,36 +251,44 @@ class LakeFSFileSystem(AbstractFileSystem):
     def info(self, path: str, **kwargs: Any) -> dict[str, Any]:
         path = self._strip_protocol(path)
 
-        # input path is a directory name
-        if path.endswith("/"):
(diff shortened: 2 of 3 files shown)


tests/test_ls.py
```diff
@@ -55,7 +55,7 @@ def test_ls_stale_cache_entry(
     lpath = str(random_file)
     rpath = f"{repository}/{temp_branch}/data/{random_file.name}"
 
-    fs.put_file(lpath, rpath)
+    fs.put_file(lpath, rpath, precheck=False)
```
nicholasjng, line 58: This is necessary now since otherwise on the checksum pre-check, `info` is called on a non-existent file, which causes the fall-through and thus another `list_objects` call.

Took me a while to figure out. Verifying the number of expected API counts is nice, but does not tell you where stuff goes wrong unfortunately.

nicholasjng: Correct.

PR #161, opened 2023-11-23
Fix `fs.get()` and `fs.put()` by expanding path with versioning prefix

nicholasjng:
Otherwise, parsing the lakeFS URI (with or without scheme) fails since the repository and ref were stripped.

Also, since the directories are part of the expanded path list by default in the base implementation, which is designed to make a dir on an `lpath` directory input in `put_file`/`get_file`, we need to check that any input to our custom implementations does not trigger the `precheck` checksum verification branch. This is done by checking if the lpath input is a directory.

PENDING: Tests (signing off now I promise, just wanted to unblock the docs)

### DIFF
--- a/src/lakefs_spec/spec.py
+++ b/src/lakefs_spec/spec.py
@@ -115,6 +115,22 @@ class LakeFSFileSystem(AbstractFileSystem):
         self.create_branch_ok = create_branch_ok
         self.source_branch = source_branch
 
+    def expand_path(
+        self,
+        path: str | list[str],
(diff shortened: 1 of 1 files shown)


nicholasjng:
An equally valid solution is to add the mentioned versioning prefix to the output of the `stat` object returned from `info`, i.e. to all constructed outputs from `stat_object`, and all outputs from `ls`. I have nothing against that personally, let me know which one is preferred.

nicholasjng:
Superseded by #166.

PR #167, opened 2023-11-24
Docs: Add third-party integrations user guide

AdrianoKF:
This PR adds a page on the integration of lakefs-spec with third-party libraries to the user guide as `docs/guide/integrations.md`.

It includes explanations and code examples for:

- Pandas
- DuckDB
- Polars
- PyArrow

> [!NOTE]
> I wonder if the code examples should be included from external Python files, just as for the quickstart guide? This would allow us to run them as part of an integration test suite for the docs.


janwillemkl (approved):
Looks good! The PyArrow example is a nice addition 😄

Issue #121, opened 2023-10-30
Improve public API docs coverage


AdrianoKF:
In order to make the generated API docs (more) useful, the code should include docstrings for (at a minimum):

- All modules: otherwise, subpages in the API reference will appear empty (since the are generated from the module docstring)
- All public members (function, classes): e.g., the `config` module is entire undocumented
- Any missing type annotations

AdrianoKF:
Closed by #168, #181, #184

PR #168, opened 2023-11-24
Add doc-strings to the client_helpers

maxmynter:
Add doc strings such that the `client_helpers` API are included in the documentation. 

This addresses https://github.com/aai-institute/lakefs-spec/issues/121

The question remains open whether we want doc-strings (and hence documentation) on all the members where I added them. Some might not be intended as public API. Refer to the linked issue for more details.

### DIFF
--- a/src/lakefs_spec/client_helpers.py
+++ b/src/lakefs_spec/client_helpers.py
@@ -35,7 +35,8 @@ def commit(
     message: str,
     metadata: dict[str, str] | None = None,
 ) -> Commit:
-    """Creates a new commit of all uncommitted changes on the branch in the lakeFS file storage
+    """
+    Creates a new commit of all uncommitted changes on the branch in the lakeFS file storage
 
     Parameters
     -------
@@ -71,7 +72,8 @@ def commit(
 def create_branch(
     client: LakeFSClient, repository: str, name: str, source_branch: str, exist_ok: bool = True
 ) -> str:
-    """Creates a branch in a lakeFS repository.
+    """
+    Creates a branch in a lakeFS repository.
 
     Parameters
     ----------
@@ -112,7 +114,8 @@ def create_branch(
 def create_repository(
     client: LakeFSClient, name: str, storage_namespace: str, exist_ok: bool = True
 ) -> Repository:
-    """Creates a new repository in the lakeFS file storage system with a specified name and storage namespace.
+    """
+    Creates a new repository in the lakeFS file storage system with a specified name and storage namespace.
 
     Parameters
     ----------
@@ -150,7 +153,8 @@ def create_repository(
 def create_tag(
     client: LakeFSClient, repository: str, ref: str | Commit, tag: str, exist_ok: bool = True
 ) -> Ref:
-    """Creates a new tag in the specified repository in the lakeFS file storage system.
+    """
+    Creates a new tag in the specified repository in the lakeFS file storage system.
 
     Parameters
     ----------
@@ -188,7 +192,8 @@ def create_tag(
 
 
 def list_tags(client: LakeFSClient, repository: str) -> list[Ref]:
-    """Lists all the tags in the specified repository in the lakeFS file storage system.
+    """
+    Lists all the tags in the specified repository in the lakeFS file storage system.
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


src/lakefs_spec/client_helpers.py
```diff
@@ -43,8 +62,7 @@ def commit(
 def create_branch(
     client: LakeFSClient, repository: str, name: str, source_branch: str, exist_ok: bool = True
 ) -> str:
-    """
-    Create a branch in a lakeFS repository.
+    """Create a branch in a lakeFS repository.
```
AdrianoKF, line 65: For consistent style within this module:

```suggestion
    """Creates a branch in a lakeFS repository.
```

AdrianoKF (commented):
Can you also add a short module docstring?

janwillemkl:
For the parameter descriptions, there is some inconsistency:

For example:

* **The** lakeFS client object
* Repository name
* **A** string representing the commit SHA

In principle it is also not needed to include the type in the description, because that information is already present in definition.

src/lakefs_spec/client_helpers.py
```diff
@@ -26,6 +35,25 @@ def commit(
     message: str,
     metadata: dict[str, str] | None = None,
 ) -> Commit:
+    """Creates a new commit of all uncommitted changes on the branch in the lakeFS file storage
```
nicholasjng, line 38: Please end with a full stop (".").

Also, do we break the line after the multiline string """? I always like to do this since it's easier to read.

src/lakefs_spec/client_helpers.py
```diff
@@ -26,6 +35,25 @@ def commit(
...
+    client: LakeFSClient
+        lakeFS client object.
+    repository: str
+        Repository name.
+    branch: str
+        Branch name.
+    message: str
+        Commit message.
+    metadata: dict[str, str] | None, optional
+        Additional metadata for the commit. Defaults to None.
```
nicholasjng, line 51:
```suggestion
        Additional metadata to include in the commit.
```

(The default is clear from the signature.)

src/lakefs_spec/client_helpers.py
```diff
@@ -26,6 +35,25 @@ def commit(
...
+
+    Returns
+    -------
+        Commit: Commit object of the lakeFS-SDK.
```
nicholasjng, line 55:
```suggestion
        Commit: The created commit object returned by the lakeFS server.
```

nicholasjng (commented):
Before I continue reviewing I want to ask:

* What is the policy on the introductory sentence in the docstring re: linebreak after the multiline intro `"""`?
* Do we want to explicitly mention what the default is in argument descriptions, even if it's in the signature?
* What tense / grammatical case are we putting the synopses into? I would argue for imperative (example: `Merge branch xyz`) instead of 3rd person (`Merges branch xyz`).
* Do we give names and types to the return values in the `Returns` section? I think we just went with a one sentence-description so far (names are not really valuable anyway, since the user will assign the name by himself).

AdrianoKF:
I took the liberty of answering a few of your questions by means of pushing a reformatting of the docstrings 😅  (without seeing your comments)

>     * What is the policy on the introductory sentence in the docstring re: linebreak after the multiline intro `"""`?
> 
>     * Do we want to explicitly mention what the default is in argument descriptions, even if it's in the signature?
Arguable, but I lean towards omitting them (if the rendered docs show the defaults correctly - which it seems they do).

>     * What tense / grammatical case are we putting the synopses into? I would argue for imperative (example: `Merge branch xyz`) instead of 3rd person (`Merges branch xyz`).
I don't care either way, but agree with what the [Google Python style guide](https://google.github.io/styleguide/pyguide.html) has to say on the matter:

> The docstring may be descriptive-style ("""Fetches rows from a Bigtable.""") or imperative-style ("""Fetch rows from a Bigtable."""), but the style should be consistent within a file.

>     * Do we give names and types to the return values in the `Returns` section? I think we just went with a one sentence-description so far (names are not really valuable anyway, since the user will assign the name by himself).

No names, IMO (in fact I edited them out)

<hr>

A few general remarks, @maxmynter:

- Sections in the Numpy docstring style start unintended (e.g., the `Notes` block would otherwise render as a markdown code block instead)
- No colons after single line identifiers (e.g., return types, exceptions), these show up in the output otherwise

I also added a `watch` block to `mkdocs.yml` for the source folder to make editing docstrings easier (since `mkdocs serve` will automatically rebuild the docs when saving a changed source file).

AdrianoKF:
Merging now, as these can always be improved later on.

PR #169, opened 2023-11-24
Allow leading `lakefs://` scheme in input paths

### LINKED ISSUE #159: Consider handling of rpaths with URI scheme included
s3fs for one parses the bucket and resource names after stripping the protocol, so maybe calling `_strip_protocol` in the `parse` routine is the way to go here.

_Originally posted by @nicholasjng in https://github.com/aai-institute/lakefs-spec/pull/156#discussion_r1402028709_

nicholasjng:
Since the scheme is a part of the URI, it should be allowed in the parsing regex. This is achieved by adding an optional positive lookahead capture group.

This has the benefit that the scheme, since we don't need it, does not appear in the regex match groups, so all other code can stay unchanged.

Adds a regression test for scheme parsing.

Fixes #159.

### DIFF
--- a/src/lakefs_spec/util.py
+++ b/src/lakefs_spec/util.py
@@ -39,10 +39,12 @@ def parse(path: str) -> tuple[str, str, str]:
     # https://docs.lakefs.io/understand/model.html#repository
     # Second regex is the branch: Only letters, digits, underscores
     # and dash, no leading dash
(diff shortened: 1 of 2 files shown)


AdrianoKF (approved):
Description nitpick: the `(?:lakefs://)?` is an optional non-capturing group, not a lookahead (`?=`) ;)

Otherwise, lgtm.

PR #170, opened 2023-11-27
Documentation cleanup [ENG-198]

AdrianoKF:
This PR should bring the documentation to a release-ready state (except the tutorial notebook, which is in scope for #158).

Major changes:

- Consistent spelling: lakeFS-spec, lakeFS, and fsspec
- Removed use cases page from docs, needs major overhaul
- Removed reference to constructor args in configuration user guide, since they can cause consistency issues

### DIFF
--- a/docs/guides/filesystem-usage.md
+++ b/docs/guides/filesystem-usage.md
@@ -15,7 +15,7 @@ For example, the URI `repo/main/data/` (note the trailing slash) refers to the `
 ## On staged versus committed changes
 
 When uploading, copying, or removing files or directories from a branch, those removal operations will result in staged changes in the repository until a commit is created.
-`lakefs-spec` does not create these commits automatically, since it separates file operations from versioning operations rigorously.
+lakeFS-spec does not create these commits automatically, since it separates file operations from versioning operations rigorously.
 If you want to conduct versioning operations, like creating commits, between file transfers, the best way to do so is by using [filesystem transactions](transactions.md).
 
 ## How to use lakeFS file system APIs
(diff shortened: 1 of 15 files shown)


nicholasjng:
>     * Removed reference to constructor args in configuration user guide, since they can cause consistency issues

Consistency of what, if I may ask? (I still agree that the section was unnecessary)

nicholasjng (approved):
Just for my understanding, what prompted the move away from backticks for lakeFS-spec and fsspec? (I was assuming that if it's a Python library, use its PyPI name in backticks)

AdrianoKF:
> > ```
> > * Removed reference to constructor args in configuration user guide, since they can cause consistency issues
> > ```
> 
> Consistency of what, if I may ask? (I still agree that the section was unnecessary)

This was in regard to what Jan had looked into, keeping the config consistent between high-level fsspec users (Pandas, ...) and transactions (in the presence of the instance cache) when using constructor args seems like a minefield.

docs/guides/filesystem-usage.md
```diff
@@ -66,7 +67,7 @@ fs.put("dir", "my-repo/my-ref/dir", recursive=True)
     fs.put_file("my-repo/my-ref/file.txt", "file.txt", use_blockstore=True)
     ```
 
-    Direct lakeFS blockstore uploads require the installation of the corresponding `fsspec` file system implementation through `pip`.
```
maxmynter, line 69: Quick Question: What is our rationale as to whether to put a tool in the backticks (``) or not?

AdrianoKF:
> Just for my understanding, what prompted the move away from backticks for lakeFS-spec and fsspec? (I was assuming that if it's a Python library, use its PyPI name in backticks)

We debated a bit, but came to the understanding that usually when you think about a product as such, names tend to be without backticks (only use them when specifically referring to a package name), think about Pandas, Numpy, ... - you don't usually see them in backticks. It also takes away quite a bit of visual clutter from the docs, especially in headings.

maxmynter (approved):

docs/guides/filesystem-usage.md
AdrianoKF: Rule of thumb was that whenever the name in the sentence refers to the product/tool itself as a whole, we don't put it in backticks.

When it specifically refers to a package name (as in the next sentence here), or a command to be run, we use backticks.

Here, `pip` kinda slipped through the cracks, but I think it could go either way, since it alludes to the command to be invoked.

PR #171, opened 2023-11-27
Add transaction usage document [ENG-150]

nicholasjng:
Mentions the most important features of the lakeFS transaction, and states that they are the intended place for versioning operations.

### DIFF
--- a/docs/guides/transactions.md
+++ b/docs/guides/transactions.md
@@ -8,10 +8,6 @@ For more information on fsspec transactions, see the official [documentation](ht
 
 The main features of the lakeFS file system transaction are:
 
-## Thread safety
-
-Through its use of `collections.deque` as a store for uploads, upload queueing and file transfers are thread-safe.
-
 ## Atomicity
 
 If an exception occurs anywhere during the transaction, all queued file uploads and versioning operations are discarded:
@@ -61,7 +57,7 @@ The full list of supported lakeFS versioning operations:
     All of the operations above are deferred, and their results are not available until completion of the transaction.
     For example, the `sha` return value of `tx.commit` will be a placeholder for the actual commit SHA computed by the lakeFS server on commit creation.
 
-    While you can use some values (branch/tag names) returned by transaction versioning helpers, we strongly advise not to reuse values outside of the transaction, since they might result in unexpected behavior.
+    While you can directly use some values (branch/tag names) returned by transaction versioning helpers, care needs to be taken with computed objects like commit SHAs:
 
     ```python
     with fs.transaction as tx:
@@ -70,7 +66,26 @@ The full list of supported lakeFS versioning operations:
     
     # This will not work: `sha` is of type `Placeholder[Commit]`
     fs.get_file(f"repo/{sha}/my-file.txt", "my-new-file.txt")
-
-    # Instead, you can use `Placeholder.unwrap()`:
-    fs.get_file(f"repo/{sha.unwrap()}/my-file.txt", "my-new-file.txt")
     ```
+    
+    See the subsection directly below on how to reuse commits created during transactions. 
+
+### A note on reusing resources created in transactions
+
+Some transaction versioning helpers create new objects in the lakeFS instance that are not known before said helpers are actually executed.
+An example of this is a commit SHA, which is only available once created by the lakeFS server.
+In the above example, a commit is created directly after a file upload, but its actual SHA identifier will not be available until the transaction is complete.
+To reuse the value later in your code, you can call `unwrap()` on the resulting [`Placeholder`](../reference/lakefs_spec/transaction.md#lakefs_spec.transaction.Placeholder) object:
+
+```python
+with fs.transaction as tx:
+    tx.put_file("my-file.txt", "repo/branch/my-file.txt")
+    sha = tx.commit("repo", "branch", message="Add my-file.txt")
+
+# Obtain the SHA value by unwrapping the placeholder first.
+fs.get_file(f"repo/{sha.unwrap()}/my-file.txt", "my-new-file.txt")
+```
+
+## Thread safety
+
+Through its use of `collections.deque` as a store for uploads, upload queueing and file transfers are thread-safe.
--- a/docs/guides/transactions.md
+++ b/docs/guides/transactions.md
@@ -68,9 +68,9 @@ The full list of supported lakeFS versioning operations:
     fs.get_file(f"repo/{sha}/my-file.txt", "my-new-file.txt")
     ```
     
-    See the subsection directly below on how to reuse commits created during transactions. 
+    See the following section on how to reuse commits created during transactions. 
 
-### A note on reusing resources created in transactions
+### Reusing resources created in transactions
 
 Some transaction versioning helpers create new objects in the lakeFS instance that are not known before said helpers are actually executed.
 An example of this is a commit SHA, which is only available once created by the lakeFS server.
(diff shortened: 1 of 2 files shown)


docs/guides/transactions.md
```diff
@@ -1,6 +1,56 @@
...
+
+The main features of the lakeFS file system transaction are:
+
+## Thread safety
+
+Through its use of `collections.deque` as a store for uploads, upload queueing and file transfers are thread-safe.
+
+## Atomicity
+
+If an exception occurs anywhere during the transaction, all queued file uploads are discarded:
```
AdrianoKF, line 16: > all queued file uploads

What about the versioning ops? It's listed in the explanation of the code example below, but I believe this should be mentioned more prominently.

docs/guides/transactions.md
```diff
@@ -1,6 +1,56 @@
...
 
-TODO: Add content
+A transaction is basically a context manager that collects all file uploads, defers them, and executes the uploads on completion of the transaction.
+They are an "all or nothing" proposition: If an error occurs during the transaction, none of the queued files are uploaded.
+
+The main features of the lakeFS file system transaction are:
+
+## Thread safety
+
+Through its use of `collections.deque` as a store for uploads, upload queueing and file transfers are thread-safe.
```
AdrianoKF, line 12: A bit too much implementation detail for my liking, then again, there's not much more to say on the topic here. (I could even see leaving out the section altogether and just sticking "thread-safe" somewhere in the introduction)

docs/guides/transactions.md
```diff
@@ -1,6 +1,56 @@
...
+
+with fs.transaction as tx:
+    fs.put_file("my-file.txt", "repo/main/my-file.txt")
+    tx.commit("repo", "main", message="Add my-file.txt")
+    raise ValueError("oops!")
+```
+
+The above code will not produce a commit on `main`, since the `ValueError` prompts a discard of the full upload queue. 
+
+## Versioning helpers
```
AdrianoKF, line 31: Maybe include a full list of the supported operations in this section?

docs/guides/transactions.md
```diff
@@ -1,6 +1,56 @@
...
+```
+
+!!! Warning
+
+    All of the operations above are deferred, and their results are not available until completion of the transaction.
+    For example, the `sha` return value of `tx.commit` will be a placeholder for the actual commit SHA computed by the lakeFS
+    server on commit creation.
+
+    While you can use some values (branch/tag names) returned by transaction versioning helpers, we strongly
+    advise not to reuse values computed in transactions since they might result in unexpected behavior.
```
AdrianoKF, line 56: > reuse

Not quite sure what this is referring to

AdrianoKF (commented):
One small request for the sake of reviewing/editing: Having multiple sentences on one line in Markdown is a bit inconvenient, since it's not quite clear in the review which sentence a comment might apply to.

Since Markdown automatically continues the same paragraph when you add a newline after a sentence, it's best (for reviewing and diffs) to put each sentence on a separate line.

docs/guides/transactions.md
nicholasjng: This is referring to the fact that `sha = tx.commit(...)` will return a `Placeholder[Commit]`, which will not work if passed to a non-transaction function later. I thought this was worth mentioning, since the following seems quite likely to happen eventually:

```python
with fs.transaction as tx:
    tx.put_file("my-file.txt", "repo/branch/my-file.txt")
    sha = tx.commit("repo", "branch", message="Add file xyz")

...
fs.get_file("repo/{sha}/file.txt", "my-new-file.txt")  # <- bang: sha is Placeholder[Commit]
```

What do you think?

AdrianoKF: Ah, that makes total sense (also, that issue would be addressed if we turn the `Placeholder` into a transparent wrapper - at least for accessing the value outside of the current transaction). I'd include the example you gave in the docs for clarity and maybe change "reuse" to "reuse outside of the transaction" (because that is the unexpected bit, I'd say)

docs/guides/transactions.md
nicholasjng: Done in [e25c6bf](https://github.com/aai-institute/lakefs-spec/pull/171/commits/e25c6bf2fc0a46929721b0dafcec76c6e02dc5b7).

docs/guides/transactions.md
nicholasjng: Mentioned the versioning operations in [e25c6bf](https://github.com/aai-institute/lakefs-spec/pull/171/commits/e25c6bf2fc0a46929721b0dafcec76c6e02dc5b7).

docs/guides/transactions.md
```diff
@@ -1,6 +1,56 @@
 # Using transactions on the lakeFS file system
 
-A transaction defers file transfers and versioning operations to a queue, which is unwound sequentially on completion.
-Transactions are thread-safe and atomic, meaning that a single failure during any transaction function causes the entire transaction to abort.
+In addition to file operations, you can carry out versioning operations in your Python code using file system *transactions*.
 
-TODO: Add content
+A transaction is basically a context manager that collects all file uploads, defers them, and executes the uploads on completion of the transaction.
```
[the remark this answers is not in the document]
nicholasjng: Added the link in [e25c6bf](https://github.com/aai-institute/lakefs-spec/pull/171/commits/e25c6bf2fc0a46929721b0dafcec76c6e02dc5b7).

docs/guides/transactions.md
AdrianoKF: Sweet, that adds a lot of value already. Can you add links to the API reference as well, please? The following URL style works in MkDocs:

```
[`commit`](../reference/lakefs_spec/transaction.md#lakefs_spec.transaction.LakeFSTransaction.commit)
```

docs/guides/transactions.md
```diff
@@ -1,6 +1,76 @@
...
+* `tag`, for creating a tag pointing to a commit.
+
+!!! Warning
+
+    All of the operations above are deferred, and their results are not available until completion of the transaction.
+    For example, the `sha` return value of `tx.commit` will be a placeholder for the actual commit SHA computed by the lakeFS server on commit creation.
+
+    While you can use some values (branch/tag names) returned by transaction versioning helpers, we strongly advise not to reuse values outside of the transaction, since they might result in unexpected behavior.
+
+    ```
```
AdrianoKF, line 66:
```suggestion
    ```python
```

docs/guides/transactions.md
AdrianoKF: Maybe let's just move the section further towards the end of the page. While I agree it's useful information, it doesn't really help the reader right away (as the other sections do).

docs/guides/transactions.md
```diff
@@ -1,6 +1,76 @@
...
+    with fs.transaction as tx:
+        tx.put_file("my-file.txt", "repo/branch/my-file.txt")
+        sha = tx.commit("repo", "branch", message="Add my-file.txt")
+    
+    # This will not work: `sha` is of type `Placeholder[Commit]`
+    fs.get_file(f"repo/{sha}/my-file.txt", "my-new-file.txt")
+
+    # Instead, you can use `Placeholder.unwrap()`:
```
AdrianoKF, line 74: I think this bit of information should go in the running text as well, since it's easy enough to miss here.

AdrianoKF (approved):

PR #177, opened 2023-11-28
Add HTTP409 check in `create_repository` for forward compatibility

### LINKED ISSUE #173: Newest `lakefs` breaks repository creation test
TL,DR: The repository creation error code is now also 409 (yay!), which is a departure from the behavior of v1.0.0.

Reproducible by testing on current main against e.g. a `treeverse/lakefs:latest` Docker container.

Error:

```
>           raise ApiException(http_resp=r)
E           lakefs_sdk.exceptions.ApiException: (409)
E           Reason: Conflict
E           HTTP response headers: HTTPHeaderDict({'Content-Type': 'application/json', 'X-Content-Type-Options': 'nosniff', 'X-Request-Id': '48f1ed1d-4c51-440c-bb43-85edb9f2c401', 'Date': 'Mon, 27 Nov 2023 20:21:48 GMT', 'Content-Length': '52'})
E           HTTP response body: {"message":"error creating repository: not unique"}
```

The fix is as easy as changing the expected response code to 409 as well as the error text.

nicholasjng:
With lakeFS 1.2.0. This breakage was observed in a local test against `treeverse/lakefs:latest`, which emits HTTP409 on an existing repo instead of HTTP400.

The code assumes that HTTP400 might happen for reasons other than an existing repository, but HTTP409 does not.

Since there is no easy way to obtain the version from the backend, we end up with an if-branch on the status code.

Fixes #173.

### DIFF
--- a/src/lakefs_spec/client_helpers.py
+++ b/src/lakefs_spec/client_helpers.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import json
 import logging
 
 from lakefs_sdk import BranchCreation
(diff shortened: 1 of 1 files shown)


nicholasjng:
It should be mentioned that there is in fact an easy way to obtain the backend version: https://pydocs-sdk.lakefs.io/docs/InternalApi.html#get_lake_fs_version

No idea about how stable it is though, since the name is `InternalApi`. It's also overkill to ask the backend what version it is before performing resource creation, so I think this is the best we can do.

Issue #165, opened 2023-11-23
`create_tag` on existing tag behavior


janwillemkl:
I find the behavior of the `exists_ok` a little confusing/misleading. 

When trying to tag another commit with an already existing tag, it silently does nothing and keeps the tag on the old commit.

I expected that the tag would be updated to reference the new commit.

I guess the question here is: Are tags immutable? If so, should we set the default to `exist_ok: bool = False`? Also this function needs to get documentation.

If we consider tags mutable, we have to update the `create_tag` function.

```python

def create_tag(
    client: LakeFSClient, repository: str, ref: str | Commit, tag: str, exist_ok: bool = True
) -> Ref:
    if isinstance(ref, Commit):
        ref = ref.id
    tag_creation = TagCreation(id=tag, ref=ref)

    try:
        return client.tags_api.create_tag(repository=repository, tag_creation=tag_creation)
    except ApiException as e:
        if e.status == 409 and exist_ok:
            return client.tags_api.get_tag(repository=repository, tag=tag). # <-- This I did not expect
        raise e
```

maxmynter:
Whether or not to allow for tag mutability depends on how we (or lakeFS) encourages to work with them.  [This](https://docs.lakefs.io/understand/model.html#tags) is their examples and description of tags:

```
Tags are a way to give a meaningful name to a specific commit. 
Using tags allow users to reference specific releases, experiments or versions by using a human friendly name.

Example tags:

v2.3 to mark a release.
dev:jane-before-v2.3-merge to mark Jane’s private temporary point.
```

We definitely want to avoid that Data Scientist B changes a tag which breaks the reproducibility of Data Scientist A's code. 

Therefore, I think it makes sense to have the tags immutable. There also exists a `delete_tag` method in the `tags_api`. Leveraging that, the reassignment is possible by deleting and subsequently creating a tag pointing to a different ref. 

This approach would make data scientists also more aware/considerate that their changes potentially break code of others. 

I will create a PR for that.

PR #179, opened 2023-11-28
Adapt `create_tag` client helper to reflect tag immutability

maxmynter:
Tags are immutable as per design from lakeFS (stated on lakeFS GUI tag site, no rename_tag API and their [docs](https://docs.lakefs.io/understand/model.html#tags)). The client helper returns the tag that is still pointing to the previously assigned commit. 

Therefore, 
- [x] set the `exist_ok` default of the `create_tag` client helper to `False` such that allowing the tag to exist (for e.g. idempotency of a script) is a conscious choice.
- [x] Log a warning that the tag was not reassigned if the tag already existed, to inform the user. 
- [x] Adapt the tutorial notebook to these changes
- [x] Add `delete_tag`  client helper such that explicit renaming by deletion and recreation is possible via our API
- [x] Create `delete_tag` tests 

Closes https://github.com/aai-institute/lakefs-spec/issues/165

- - -
This change is [Reviewable](https://reviewable.io/reviews/aai-institute/lakefs-spec/179)

### DIFF
--- a/docs/tutorials/demo_data_science_project.ipynb
+++ b/docs/tutorials/demo_data_science_project.ipynb
@@ -556,7 +556,7 @@
     "\n",
     "The above method for data versioning works great when you have experiment tracking tools to store and retrieve the commit SHA in automated pipelines. But it can be tedious to retrieve in manual prototyping. We can make selected versions of the dataset more accessible with semantic versioning by attaching a human-interpretable tag to a specific commit SHA.\n",
     "\n",
-    "Creating a tag is easiest when done inside a transaction, just like the files we already uploaded. To do this, simply call `tx.tag` on the transaction and supply the repository name, the commit SHA to tag, and the intended tag name."
+    "Creating a tag is easiest when done inside a transaction, just like the files we already uploaded. To do this, simply call `tx.tag` on the transaction and supply the repository name, the commit SHA to tag, and the intended tag name. The `exist_ok` flag handles creation errors, thrown if the tag does already exist. Since tags are immutable, the returned tag still points to the commit SHA it was previously assigned to. "
    ]
   },
   {
@@ -570,7 +570,7 @@
    "source": [
     "with fs.transaction as tx:\n",
     "    # the `tag` result is simply the tag name, in this case 'train-test-split-2010'.\n",
-    "    tag = tx.tag(repository=REPO_NAME, ref=fixed_commit_id, tag='train-test-split-2010')"
+    "    tag = tx.tag(repository=REPO_NAME, ref=fixed_commit_id, tag='train-test-split-2010', exist_ok=True)"
    ]
   },
   {
--- a/docs/tutorials/demo_data_science_project.ipynb
+++ b/docs/tutorials/demo_data_science_project.ipynb
@@ -556,7 +556,7 @@
     "\n",
     "The above method for data versioning works great when you have experiment tracking tools to store and retrieve the commit SHA in automated pipelines. But it can be tedious to retrieve in manual prototyping. We can make selected versions of the dataset more accessible with semantic versioning by attaching a human-interpretable tag to a specific commit SHA.\n",
     "\n",
-    "Creating a tag is easiest when done inside a transaction, just like the files we already uploaded. To do this, simply call `tx.tag` on the transaction and supply the repository name, the commit SHA to tag, and the intended tag name. The `exist_ok` flag handles creation errors, thrown if the tag does already exist. Since tags are immutable, the returned tag still points to the commit SHA it was previously assigned to. "
+    "Creating a tag is easiest when done inside a transaction, just like the files we already uploaded. To do this, simply call `tx.tag` on the transaction and supply the repository name, the commit SHA to tag, and the intended tag name. Tags are immutable once created, so attempting to tag two different commits with the same name will result in an error."
    ]
   },
   {
--- a/docs/tutorials/demo_data_science_project.ipynb
+++ b/docs/tutorials/demo_data_science_project.ipynb
@@ -570,7 +570,7 @@
    "source": [
     "with fs.transaction as tx:\n",
     "    # the `tag` result is simply the tag name, in this case 'train-test-split-2010'.\n",
-    "    tag = tx.tag(repository=REPO_NAME, ref=fixed_commit_id, tag='train-test-split-2010', exist_ok=True)"
+    "    tag = tx.tag(repository=REPO_NAME, ref=fixed_commit_id, tag='train-test-split-2010')"
    ]
   },
   {
--- a/src/lakefs_spec/client_helpers.py
+++ b/src/lakefs_spec/client_helpers.py
@@ -193,6 +193,21 @@ def create_tag(
             logger.warning(f"The tag, '{tag}' already exists. It was not reassigned.")
             return client.tags_api.get_tag(repository=repository, tag=tag)
         raise e
+    
+def delete_tag(client: LakeFSClient, repository: str, tag:str) -> None:
+    """
+    Delete the specified tag from a repository.
+
+    Parameters
+    ----------
+    client : LakeFSClient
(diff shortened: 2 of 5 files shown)


src/lakefs_spec/client_helpers.py
```diff
@@ -190,10 +190,27 @@ def create_tag(
         return client.tags_api.create_tag(repository=repository, tag_creation=tag_creation)
     except ApiException as e:
         if e.status == 409 and exist_ok:
+            logger.warning(f"The tag, '{tag}' already exists. It was not reassigned.")
```
nicholasjng, line 193: Make it short and simple, like `f"tag {tag} already exists and is immutable"`, and then match logs against `"tag .* already exists`.

nicholasjng (commented):
I'm not sure what we settled on to fix the issue, but what do you say to the following behavior:

On HTTP409 (tag already exists): Get tag, check its `commit_id`, and verify it matches `rev_parse(client, repo, ref)`, otherwise, hard error.

This removes the ambiguity (and the need for `exist_ok`) entirely, and the fewest users will probably be saying "I want this tag and I don't care where it points", or am I mistaken?

So, in other words, allow existence ONLY if name and revision match.

docs/tutorials/demo_data_science_project.ipynb
```diff
@@ -556,7 +556,7 @@
     "\n",
     "The above method for data versioning works great when you have experiment tracking tools to store and retrieve the commit SHA in automated pipelines. But it can be tedious to retrieve in manual prototyping. We can make selected versions of the dataset more accessible with semantic versioning by attaching a human-interpretable tag to a specific commit SHA.\n",
     "\n",
-    "Creating a tag is easiest when done inside a transaction, just like the files we already uploaded. To do this, simply call `tx.tag` on the transaction and supply the repository name, the commit SHA to tag, and the intended tag name."
+    "Creating a tag is easiest when done inside a transaction, just like the files we already uploaded. To do this, simply call `tx.tag` on the transaction and supply the repository name, the commit SHA to tag, and the intended tag name. The `exist_ok` flag handles creation errors, thrown if the tag does already exist. Since tags are immutable, the returned tag still points to the commit SHA it was previously assigned to. "
```
nicholasjng, line 559: Since `exist_ok` is now out of scope, I would propose this:

```suggestion
    "Creating a tag is easiest when done inside a transaction, just like the files we already uploaded. To do this, simply call `tx.tag` on the transaction and supply the repository name, the commit SHA to tag, and the intended tag name. Tags are immutable once created, so attempting to tag two different commits with the same name will result in an error."
```

src/lakefs_spec/client_helpers.py
```diff
@@ -189,11 +185,29 @@ def create_tag(
     try:
         return client.tags_api.create_tag(repository=repository, tag_creation=tag_creation)
     except ApiException as e:
-        if e.status == 409 and exist_ok:
-            return client.tags_api.get_tag(repository=repository, tag=tag)
+        target_commit = rev_parse(client=client, repository=repository, ref=ref)
+        existing_tag = client.tags_api.get_tag(repository=repository, tag=tag)
+        if e.status == 409 and existing_tag.commit_id == target_commit.id:
+            return existing_tag
```
nicholasjng, line 191: For performance reasons, please move the rev-parse attempt in the `if e.status == 409` block so that it doesn't execute on e.g. a `PermissionError`:

```suggestion
        if e.status == 409:
                tagged_commit = rev_parse(client=client, repository=repository, ref=ref)
                existing_tag = client.tags_api.get_tag(repository=repository, tag=tag)
                if tagged_commit.id == existing_tag.commit_id:
                    return existing_tag
```

src/lakefs_spec/client_helpers.py
```diff
@@ -189,11 +185,29 @@ def create_tag(
...
         raise e
 
 
+def delete_tag(client: LakeFSClient, repository: str, tag: str) -> None:
```
nicholasjng, line 195: Should we support `Ref` or `Commit` instances as tag inputs here, like e.g. in `rev_parse`?

tests/test_client_helpers.py
```diff
@@ -4,37 +4,84 @@
...
-        existing_tag = client_helpers.create_tag(
-            client=fs.client, repository=repository, ref=temp_branch, tag=tag
+        commit_random_file_on_branch(
+            random_file_factory=random_file_factory,
+            fs=fs,
+            repository=repository,
+            temp_branch=temp_branch,
         )
-        assert new_tag == existing_tag
+        with pytest.raises(ApiException):
```
nicholasjng, line 60: I assume that the error here will be a "tag already exists" error, have you verified this? What error message does this produce?

tests/util.py
```diff
@@ -82,3 +85,17 @@ def make(self, fname: Optional[str] = None, size: int = 2**10) -> Path:
...
+    random_file_factory: RandomFileFactory, fs: LakeFSFileSystem, repository: str, temp_branch: str
+) -> None:
+    random_file = random_file_factory.make()
+    lpath = str(random_file)
+    rpath = f"{repository}/{temp_branch}/{random_file.name}"
+
+    fs.put(lpath, rpath, precheck=False)
+
+    commit = client_helpers.commit(
+        client=fs.client, repository=repository, branch=temp_branch, message="Commit File Factory"
```
nicholasjng, line 100:
```suggestion
        client=fs.client, repository=repository, branch=temp_branch, message=f"Add file {random_file.name!r}"
```

See the other similar comment - can also be done in a follow-up if you prefer.

src/lakefs_spec/client_helpers.py
maxmynter: Decided on just supporting `Ref` since the `Commit` object does not contain any reference to a tag. Plus, `list_tags` returns  a list of `Ref`s. Lastly, there could be edge cases such as multiple tags to a single commit (which should we then delete) and we want to avoid someone accidentally deleting someone elses tag.

tests/test_client_helpers.py
maxmynter: Btw, 
the terminal outputs the following:
```
raise ApiException(http_resp=r)
E           lakefs_sdk.exceptions.ApiException: (409)
E           Reason: Conflict
E           HTTP response headers: HTTPHeaderDict({'Content-Type': 'application/json', 'X-Content-Type-Options': 'nosniff', 'X-Request-Id': '19a26127-d216-4e93-af9b-45aff9e6a940', 'Date': 'Tue, 28 Nov 2023 19:08:34 GMT', 'Content-Length': '45'})
E           HTTP response body: {"message":"tag already exists: not unique"}

.venv/lib/python3.11/site-packages/lakefs_sdk/rest.py:237: ApiException
```

src/lakefs_spec/client_helpers.py
```diff
@@ -189,11 +185,35 @@ def create_tag(
...
+        Name of the repository from which the tag will be deleted.
+    tag : str
+        Name of the tag to be deleted.
+    """
+    if isinstance(tag, str):
+        client.tags_api.delete_tag(repository=repository, tag=tag)
+    elif isinstance(tag, Ref):
+        client.tags_api.delete_tag(repository=repository, tag=tag.id)
+    else:
+        raise ValueError(r"Type of tag must be str or Ref.")
```
nicholasjng, line 214: This is unnecessary with a proper interface typing - after we tell users what is allowed, it's not our responsibility anymore if they pass something else.

I would go with this:

```suggestion
    if isinstance(tag, Ref):
        tag = tag.id

    client.tags_api.delete_tag(repository=repository, tag=tag.id)
```

tests/test_client_helpers.py
nicholasjng: Fair. Can you `assert e.status == 409`?

nicholasjng (approved):
Nice!

PR #180, opened 2023-11-28
Add depagination helper

### LINKED ISSUE #178: Add depagination helper
This is about a currently silent bug in most of our response handling code: API responses are currently not depaginated anywhere except in `fs.ls()`, meaning that not all resources are returned from `GET` requests if the amount of resources is very large. Since we are dealing with mostly empty repos (local ephemeral instances), we do not currently observe the effects of this.

Example: A repository has 110 tags, `client_helpers.list_tags()` returns at most 100 results -> we are missing at least 10 tags in the response.

lakeFS even mentions a solution here (https://docs.lakefs.io/integrations/python.html#usage-examples):

```python
# source: https://docs.lakefs.io/integrations/python.html#usage-examples
def pagination_helper(page_fetcher, **kwargs):
    """Helper function to iterate over paginated results"""
    while True:
        resp = page_fetcher(**kwargs)
        yield from resp.results
        if not resp.pagination.has_more:
            break
        kwargs['after'] = resp.pagination.next_offset
```

`page_fetcher` is the API that's being called, for example `objects_api.list_object`. This works out, but is entirely untyped, so we have to think a bit about how to type this properly.

Questions:

- [ ] Can we type the API?
- [ ] Can we type the kwargs to dynamically accept a Python API spec?
- [ ] Can we define a generic that enables automatic typing of the return object?

What's a little bit unfortunate is that the pydantic response models (`resp` in the above code) do not share a common base class, so we need to roll our own.

nicholasjng:
This is basically the lakeFS SDK example with some typing to improve the expected inputs and outputs.

I'm pretty sure the generic typing does not work, since mypy does not make the connection from a pydantic response model (say, `RefList`) to the result type (in this case, `list[Ref]`).

My idea was a "C++ template specialization" kind of approach, forwarding `RefList = PaginatedApiResponse[Ref]`, but it seems like that's not directly possible in the current typing.

This also likely needs some logic to accommodate Python 3.9 with the `ParamSpec` usage and so on.

Closes #178.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aai-institute/lakefs-spec/180)

### DIFF
--- a/src/lakefs_spec/util.py
+++ b/src/lakefs_spec/util.py
@@ -1,25 +1,22 @@
 import hashlib
 import re
-from typing import Callable, Generator, ParamSpec, Protocol, TypeVar
+from typing import Any, Callable, Generator, Protocol
 
+from lakefs_sdk import Pagination
 from lakefs_sdk import __version__ as __lakefs_sdk_version__
-from lakefs_sdk.models import Pagination
 
 lakefs_sdk_version = tuple(int(v) for v in __lakefs_sdk_version__.split("."))
 del __lakefs_sdk_version__
 
-T = TypeVar("T")
-P = ParamSpec("P")
 
-
-class PaginatedApiResponse(Protocol[T]):
+class PaginatedApiResponse(Protocol):
     pagination: Pagination
-    results: list[T]
+    results: list
 
 
 def depaginate(
-    api: Callable[P, PaginatedApiResponse[T]], *args: P.args, **kwargs: P.kwargs
-) -> Generator[T, None, None]:
+    api: Callable[..., PaginatedApiResponse], *args: Any, **kwargs: Any
+) -> Generator[Any, None, None]:
     """Send a number of lakeFS API response documents to a generator."""
     while True:
         resp = api(*args, **kwargs)
--- a/src/lakefs_spec/util.py
+++ b/src/lakefs_spec/util.py
@@ -1,10 +1,32 @@
 import hashlib
 import re
+from typing import Callable, Generator, ParamSpec, Protocol, TypeVar
 
 from lakefs_sdk import __version__ as __lakefs_sdk_version__
+from lakefs_sdk.models import Pagination
 
 lakefs_sdk_version = tuple(int(v) for v in __lakefs_sdk_version__.split("."))
 del __lakefs_sdk_version__
+
+T = TypeVar("T")
+P = ParamSpec("P")
+
+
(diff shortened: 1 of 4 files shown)


src/lakefs_spec/util.py
```diff
@@ -1,11 +1,33 @@
...
+P = ParamSpec("P")
+
+
+class PaginatedApiResponse(Protocol[T]):
+    pagination: Pagination
+    results: list[T]
+
+
+def depaginate(
+    api: Callable[P, PaginatedApiResponse[T]], *args: P.args, **kwargs: P.kwargs
```
AdrianoKF, line 21: I'm not sure the type hints really do us any good here. How about just using `cast(depaginate(), ...)` on the results in the callers? E.g.,

```python
objects = list(depaginate(self.client.objects_api.list_objects, repository, ref, **kwargs))
for obj in cast(list[ObjectStats], objects):
    # obj is of type ObjectStats
    ...
```

nicholasjng: Sounds good, also solves the 3.9 problems with using `ParamSpec`.

I don't think mypy can even do anything with the spec, I tried inserting a random keyword argument into the `list_objects` call and it didn't throw any error.

nicholasjng: Taken over in [5a5a1dd](https://github.com/aai-institute/lakefs-spec/pull/180/commits/5a5a1ddb15a51c411baaecb97e3bc47266da933c).

AdrianoKF: One minor thing, I believe the cast should be (if I'm not mistaken):

```python
objects = depaginate(self.client.objects_api.list_objects, repository, ref, **kwargs)
for obj in cast(Iterable[ObjectStats], objects):
```

Note the `Iterable`, since the generator doesn't return a `list`.

nicholasjng: Fixed in [a6661d9](https://github.com/aai-institute/lakefs-spec/pull/180/commits/a6661d99e25552041635d576ca0c67caf56e2f21).

AdrianoKF (approved):
> Reviewed 1 of 2 files at r2, 2 of 3 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/aai-institute/lakefs-spec/180)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nicholasjng)

PR #181, opened 2023-11-29
Create API docstrings for the `LakeFSFileSystem` and `LakeFSFile`

nicholasjng:
Annotates all overloaded methods, both public and private, on the `LakeFSFileSystem` and `LakeFSFile` classes.

Some hints about input values (e.g. `rpath`s having to be fully qualified lakeFS URIs) were duplicated across all APIs, since a user might use only a subset of APIs.

Also forwards the `LakeFSFile`'s keyword arguments supplied on construction to the respective `objects_api.{get,put}_object` calls.

- - -
This change is [Reviewable](https://reviewable.io/reviews/aai-institute/lakefs-spec/181)

### DIFF
--- a/src/lakefs_spec/spec.py
+++ b/src/lakefs_spec/spec.py
@@ -1,3 +1,8 @@
+"""
+Contains the core interface definitions for file system interaction with lakeFS from Python,
+namely the ``LakeFSFileSystem`` and ``LakeFSFile`` classes.
+"""
+
 from __future__ import annotations
 
 import io
@@ -12,7 +17,7 @@ from pathlib import Path
 from typing import Any, Generator, Iterable, Literal, cast
 
 from fsspec import filesystem
-from fsspec.callbacks import Callback, NoOpCallback
+from fsspec.callbacks import _DEFAULT_CALLBACK, Callback
 from fsspec.spec import AbstractBufferedFile, AbstractFileSystem
 from fsspec.utils import stringify_path
 from lakefs_sdk import Configuration
@@ -26,8 +31,6 @@ from lakefs_spec.errors import translate_lakefs_error
 from lakefs_spec.transaction import LakeFSTransaction
 from lakefs_spec.util import depaginate, md5_checksum, parse
 
-_DEFAULT_CALLBACK = NoOpCallback()
-
 logger = logging.getLogger(__name__)
 
 EmptyYield = Generator[None, None, None]
@@ -37,8 +40,7 @@ class LakeFSFileSystem(AbstractFileSystem):
     """
     lakeFS file system implementation.
 
-    The client is immutable in this implementation, so different users need different
-    file systems.
+    The client is immutable in this implementation, so different users need different file systems.
     """
 
     protocol = "lakefs"
@@ -60,7 +62,7 @@ class LakeFSFileSystem(AbstractFileSystem):
         **storage_options: Any,
     ):
         """
-        The lakeFS file system constructor.
+        Construct a lakeFS file system.
 
         Parameters
         ----------
@@ -117,10 +119,11 @@ class LakeFSFileSystem(AbstractFileSystem):
 
     @property
     def transaction(self):
-        """A context within which files are committed together upon exit
+        """
+        A context manager within which file uploads and versioning operations are deferred to a
+        queue, and carried during __exit__.
 
-        Requires the file class to implement `.commit()` and `.discard()`
-        for the normal and exception cases.
+        Requires the file class to implement `.commit()` and `.discard()` for the normal and exception cases.
         """
         self._transaction: LakeFSTransaction | None
         if self._transaction is None:
@@ -128,19 +131,48 @@ class LakeFSFileSystem(AbstractFileSystem):
         return self._transaction
 
     def start_transaction(self):
-        """Begin write transaction for deferring files, non-context version"""
+        """
+        Prepare a lakeFS file system transaction without entering the transaction context yet.
+        """
         self._intrans = True
         self._transaction = LakeFSTransaction(self)
         return self.transaction
 
     @contextmanager
     def wrapped_api_call(self, message: str | None = None, set_cause: bool = True) -> EmptyYield:
+        """
+        A context manager to wrap lakeFS API calls, translating API errors to Python-native OS errors if they occur.
+
+        Meant for internal use.
+
+        Parameters
+        ----------
+        message: str | None
+            A custom error message to emit instead of parsing the API error response.
+        set_cause: bool
+            Whether to include the original lakeFS API error in the resulting traceback.
+        """
         try:
             yield
         except ApiException as e:
             raise translate_lakefs_error(e, message=message, set_cause=set_cause)
 
     def checksum(self, path: str | os.PathLike[str]) -> str | None:
+        """
+        Get a remote lakeFS file object's checksum.
+
+        This is usually its MD5 hash, unless another hash function was used on upload.
+
+        Parameters
+        ----------
+        path: str
+            The remote path to look up the lakeFS checksum for. Must point to a single file object.
+
+        Returns
+        -------
+        str | None
+            The remote file's checksum, or None (if ``path`` points to a directory or does not exist).
+        """
         path = stringify_path(path)
         try:
             return self.info(path).get("checksum")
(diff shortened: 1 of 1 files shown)


src/lakefs_spec/spec.py
```diff
@@ -116,36 +118,88 @@ def __init__(
...
+        Prepare a lakeFS file system transaction without entering the transaction context yet.
+        """
         self._intrans = True
         self._transaction = LakeFSTransaction(self)
         return self.transaction
 
     @contextmanager
     def wrapped_api_call(self, message: str | None = None, set_cause: bool = True) -> EmptyYield:
+        """
+        A context manager to wrap lakeFS API calls, translating API errors to Python-native OS errors if they occur.
```
maxmynter, line 143:
```suggestion
        Context manager to wrap lakeFS API calls, translating API errors to Python-native OS errors.
```

src/lakefs_spec/spec.py
```diff
@@ -116,36 +118,88 @@ def __init__(
 
     @property
     def transaction(self):
-        """A context within which files are committed together upon exit
+        """
+        A context manager within which file uploads and versioning operations are deferred to a
+        queue, and carried during __exit__.
```
maxmynter, line 123:
```suggestion
        Context manager within which file uploads and versioning operations are deferred to a
        queue, and carried during __exit__.
```

src/lakefs_spec/spec.py
maxmynter: I would in general prefer omit the articles. But I'm not emotional attached to whatever we agree upon.

src/lakefs_spec/spec.py
```diff
@@ -460,13 +683,30 @@ def __init__(
             create_branch(self.fs.client, repository, branch, self.fs.source_branch)
 
     def _upload_chunk(self, final: bool = False) -> bool:
-        """Commits the file on final chunk via single-shot upload, no-op otherwise."""
+        """
+        Commits the file on final chunk via single-shot upload, no-op otherwise.
```
maxmynter, line 687:
```suggestion
        Commit the file on final chunk via single-shot upload, no-op otherwise.
```

maxmynter (approved):

src/lakefs_spec/spec.py
```diff
@@ -244,6 +344,24 @@ def info(self, path: str, **kwargs: Any) -> dict[str, Any]:
...
+            The path under which to list objects. Must be a fully qualified lakeFS URI.
+            Can also point to a file, in which case the file's metadata will be returned.
+        detail: bool
+            Whether to obtain all metadata on the requested objects or just their names.
+        kwargs: Any
+            Additional keyword arguments to pass to ``LakeFSClient.objects_api.list_objects()``.
+
+        Returns
+        -------
+        list[str | dict[str, Any]]
```
AdrianoKF, line 362: Inconsistent with the method signature (but we can tighten the type bounds on the method since it's covariant with the base method)

src/lakefs_spec/spec.py
```diff
@@ -332,6 +476,27 @@ def put_file_to_blockstore(
...
+        Parameters
+        ----------
+        lpath: str
+            The local path to upload to the lakeFS block storage.
+        rpath: str
+            The remote target path to upload the local file to. Must be a fully qualified lakeFS URI.
+        callback: fsspec.callbacks.Callback
+            An fsspec callback to use during the operation. Can be used to report download progress.
+        presign: bool
+            Whether to use pre-signed URLs to upload the object via HTTP(S) using ``urllib.request``.
```
AdrianoKF, line 496: Mention the implications if `presign=False`: must have locally available credentials for accessing the blockstore (right?)

src/lakefs_spec/spec.py
```diff
@@ -397,6 +562,29 @@ def put_file(
...
+            The remote target path to upload the local file to. Must be a fully qualified lakeFS URI.
+        callback: fsspec.callbacks.Callback
+            An fsspec callback to use during the operation. Can be used to report download progress.
+        precheck: bool
+            Check if ``lpath`` already exists and compare its checksum with that of ``rpath``, skipping the download if they match.
+        use_blockstore: bool
+            Optionally upload the file directly to the underlying block storage, thereby bypassing the lakeFS server and saving a
+            file transfer. Preferable for uploads of large files.
+        presign: bool
+            Whether to use pre-signed URLs to upload the object if ``use_blockstore == True``.
```
AdrianoKF, line 582: See above; mention the implications if `use_blockstore=True` and `presign=False`

src/lakefs_spec/spec.py
```diff
@@ -419,6 +607,16 @@ def put_file(
             super().put_file(lpath=lpath, rpath=rpath, callback=callback, **kwargs)
 
     def rm_file(self, path: str) -> None:
+        """
+        Stage a remote file for removal on a lakeFS server.
+
+        After executing this function, the requested file will not actually be removed from the requested branch until another commit is created.
```
AdrianoKF, line 613:
```suggestion
        The file will not actually be removed from the branch until a commit is created.
```

src/lakefs_spec/spec.py
```diff
@@ -442,6 +640,31 @@ def __init__(
...
+        cache_type: str
+            Cache type to use in the ``AbstractBufferedFile`` superclass.
+        cache_options: dict[str, Any] | None
+            Additional caching options to pass to the ``AbstractBufferedFile`` superclass.
+        size: int | None
+            If given and ``mode == 'rb'``, this will be used as the file size (in bytes).
+        kwargs: Any
+            Additional keyword arguments to pass to ``LakeFSClient.objects_api.get_object()`` on download (``mode == 'rb'),
+            or ``LakeFSClient.objects_api.put_object()`` on upload (``mode == 'wb').
+        """
```
AdrianoKF, line 667: The Numpy doc style puts the constructor documentation into the class docstring: https://numpydoc.readthedocs.io/en/latest/format.html#documenting-classes

    The constructor (__init__) should also be documented here, the [Parameters](https://numpydoc.readthedocs.io/en/latest/format.html#params) section of the docstring details the constructor’s parameters.

src/lakefs_spec/spec.py
```diff
@@ -442,6 +640,31 @@ def __init__(
...
+        block_size: int | str
+            The file block size to read at a time. If not set, falls back to fsspec's default blocksize of 5 MB.
+        autocommit: bool
+            Whether to write the file buffer automatically to lakeFS on file closing in write mode.
+        cache_type: str
+            Cache type to use in the ``AbstractBufferedFile`` superclass.
+        cache_options: dict[str, Any] | None
+            Additional caching options to pass to the ``AbstractBufferedFile`` superclass.
+        size: int | None
+            If given and ``mode == 'rb'``, this will be used as the file size (in bytes).
```
AdrianoKF, line 663: I think the docstring from the base method better conveys the idea (despite the grammatical error 😅):

> If given and in read mode, suppressed having to look up the file size

Maybe let's settle on

```suggestion
            If given and ``mode == 'rb'``, this will be used as the file size (in bytes) instead of determining it from the remote file.
```

src/lakefs_spec/spec.py
```diff
@@ -442,6 +640,31 @@ def __init__(
...
+            The lakeFS file system associated to this file.
+        path: str
+            The remote path to either up- or download depending on ``mode``. Must be a fully qualified lakeFS URI.
+        mode: Literal["rb", "wb"]
+            The file mode indicating its purpose. Use ``rb`` for downloads from lakeFS, ``wb`` for uploads to lakeFS.
+        block_size: int | str
+            The file block size to read at a time. If not set, falls back to fsspec's default blocksize of 5 MB.
+        autocommit: bool
+            Whether to write the file buffer automatically to lakeFS on file closing in write mode.
+        cache_type: str
```
AdrianoKF, line 658: The base class has more information on the allowed values:

> cache_type: {"readahead", "none", "mmap", "bytes"}, default "readahead"
>            Caching policy in read mode. See the definitions in ``core``.

src/lakefs_spec/spec.py
```diff
@@ -442,6 +640,31 @@ def __init__(
...
+            Cache type to use in the ``AbstractBufferedFile`` superclass.
+        cache_options: dict[str, Any] | None
+            Additional caching options to pass to the ``AbstractBufferedFile`` superclass.
```
AdrianoKF, line 661: Mention that this is going straight to the constructor of the cache:
>        cache_options : dict
>            Additional options passed to the constructor for the cache specified
>            by `cache_type`.

src/lakefs_spec/spec.py
```diff
@@ -460,13 +683,30 @@ def __init__(
...
+        Parameters
+        ----------
+        final: bool
+            Proceed with uploading the file if ``self.autocommit == True``.
+
+        Returns
+        -------
+        bool
+            If the file buffer needs more data to be written before initiating the upload.
+
```
AdrianoKF, line 698:
```suggestion
```

src/lakefs_spec/spec.py
```diff
@@ -527,8 +769,25 @@ def flush(self, force: bool = False) -> None:
             self.offset += self.buffer.seek(0, 2)
 
     def _fetch_range(self, start: int, end: int) -> bytes:
+        """
+        Fetch a byte range of the ``LakeFSFile``'s target remote path.
+
+        The byte range is right-exclusive, meaning that the amount of transferred bytes equals end - start.
```
AdrianoKF, line 775:
```suggestion
        The byte range is right-exclusive, meaning that the amount of transferred bytes equals ``end - start``.
```

Issue #127, opened 2023-11-02
Allow passing Path-like inputs as local paths


AdrianoKF:
In #123, we added type hints that allow passing other Path-like inputs besides `str` to filesystem operations. However, they are not currently handled correctly internally when passing a `Path`, e.g., in `put()`:

```python
from lakefs_spec import LakeFSFileSystem
from pathlib import Path

fs = LakeFSFileSystem()
fs.put(Path("demo.txt"), "repo/main/demo.txt")
```

```
Traceback (most recent call last):
  File "/home/adriano/tmp/poetry-test/demo.py", line 5, in <module>
    fs.put(Path("demo.txt"), "repo/main/demo.txt")
  File "/home/adriano/tmp/poetry-test/.venv/lib/python3.11/site-packages/lakefs_spec/spec.py", line 570, in put
    super().put(
  File "/home/adriano/tmp/poetry-test/.venv/lib/python3.11/site-packages/fsspec/spec.py", line 1038, in put
    lpaths = fs.expand_path(lpath, recursive=recursive, maxdepth=maxdepth)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/adriano/tmp/poetry-test/.venv/lib/python3.11/site-packages/fsspec/spec.py", line 1153, in expand_path
    path = [self._strip_protocol(p) for p in path]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: 'PosixPath' object is not iterable
```

All methods that take a `str | os.PathLike[str] | pathlib.Path` parameter should be examined and adapted to correctly work with all types of allowed inputs.

PR #182, opened 2023-11-29
Allow Path like inputs in public LakeFSFileSystem methods

maxmynter:
Closes https://github.com/aai-institute/lakefs-spec/issues/127

- [x] Implement extended path handling
- [ ] ~~Update and extend tests~~
- [ ] ~~Update Docstrings (Types)~~

- - -
This change is [Reviewable](https://reviewable.io/reviews/aai-institute/lakefs-spec/182)

### DIFF
--- a/src/lakefs_spec/spec.py
+++ b/src/lakefs_spec/spec.py
@@ -4,6 +4,7 @@ import io
 import logging
 import mimetypes
 import operator
+import pathlib
 import os
 import urllib.error
 import urllib.request
@@ -23,7 +24,7 @@ from lakefs_spec.client_helpers import create_branch
 from lakefs_spec.config import LakectlConfig
 from lakefs_spec.errors import translate_lakefs_error
 from lakefs_spec.transaction import LakeFSTransaction
-from lakefs_spec.util import FilePathType, PathHandler, depaginate, md5_checksum, parse
+from lakefs_spec.util import depaginate, md5_checksum, parse, ensure_path_is_str
 
 _DEFAULT_CALLBACK = NoOpCallback()
 
@@ -139,7 +140,7 @@ class LakeFSFileSystem(AbstractFileSystem):
         except ApiException as e:
             raise translate_lakefs_error(e, message=message, set_cause=set_cause)
 
-    def checksum(self, path: FilePathType) -> str | None:
+    def checksum(self, path: str) -> str | None:
         try:
             return self.info(path).get("checksum")
         except FileNotFoundError:
@@ -183,16 +184,17 @@ class LakeFSFileSystem(AbstractFileSystem):
 
     def get_file(
         self,
-        rpath: FilePathType,
-        lpath: FilePathType,
+        rpath: str | os.PathLike[str] | pathlib.Path,
+        lpath: str | os.PathLike[str] | pathlib.Path,
         callback: Callback = _DEFAULT_CALLBACK,
         outfile: Any = None,
         precheck: bool = True,
         **kwargs: Any,
     ) -> None:
-        rpath = PathHandler(rpath)
-        lpath = PathHandler(lpath)
-        if precheck and lpath.exists() and lpath.is_file():
+        rpath = ensure_path_is_str(rpath)
+        lpath = ensure_path_is_str(rpath)
+        lp = Path(lpath)
+        if precheck and lp.exists() and lp.is_file():
             local_checksum = md5_checksum(lpath, blocksize=self.blocksize)
             remote_checksum = self.info(rpath).get("checksum")
             if local_checksum == remote_checksum:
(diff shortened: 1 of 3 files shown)


src/lakefs_spec/util.py
```diff
@@ -25,6 +27,16 @@ def depaginate(
             break
         kwargs["after"] = resp.pagination.next_offset
 
+def ensure_path_is_str(path: Union[str, os.PathLike, pathlib.Path]) -> str:
```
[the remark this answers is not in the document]
maxmynter: @nicholasjng I just had the idea to use a utility class instead of this one helper function to handle the paths in `lakeFS-spec` and ensure we only pass strings once it leaves our domain of concern. Let me know what you think :)

src/lakefs_spec/util.py
```diff
@@ -58,8 +89,9 @@ def parse(path: str) -> tuple[str, str, str]:
     # https://docs.lakefs.io/understand/model.html#repository
     # Second regex is the branch: Only letters, digits, underscores
     # and dash, no leading dash
+    path = PathHandler(path)
```
maxmynter, line 92: Question: How do we intend to use the `parse`  (and `md5_checksum`) method in other places?
If we don't we could directly enforce a `PathHandler` type and remove the typecasting in the first line of the function.

src/lakefs_spec/util.py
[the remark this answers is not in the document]
nicholasjng: What is the benefit of this over `pathlib.Path`? `PathHandler` is not a string, but it's not a `Path`, either.

The idea is nice, but I don't see how this is immediately useful for us. I think we should stay with fsspec's style as much as possible, which seems to be "immediately coerce to string via `stringify_path`".

So it's more function-driven instead of OOP, which makes sense for the way the interface is designed. What do you think?

(I personally think that for `(put|get)_file`, this should be as easy as calling `stringify_path` on both `lpath` and `rpath` after the precheck block. The `(put|get)` methods are another story, and might require sending a patch to `fsspec` itself, see my comment on the original issue).

src/lakefs_spec/util.py
nicholasjng: I think at most, users will pass path-like objects for the `lpath` argument, as the `rpath` argument is not really that interesting to wrap into a path (the methods on it don't work for a remote path, which is what `universal_pathlib` is trying to fix).

So we can leave those interfaces as is, and just coerce directly before. (That being said, `lpath` should be a `str | Path` on `md5_checksum`.)

src/lakefs_spec/util.py
```diff
@@ -36,7 +40,7 @@ def md5_checksum(lpath: str, blocksize: int = 2**22) -> str:
     return file_hash.hexdigest()
 
 
-def parse(path: str) -> tuple[str, str, str]:
+def parse(path: Union[str, os.PathLike[str], Path]) -> tuple[str, str, str]:
```
nicholasjng, line 43: This lives behind the API boundary, so its signature should not be changed. Please revert.

src/lakefs_spec/util.py
```diff
@@ -26,7 +29,8 @@ def depaginate(
         kwargs["after"] = resp.pagination.next_offset
 
 
-def md5_checksum(lpath: str, blocksize: int = 2**22) -> str:
+def md5_checksum(lpath: Union[str, os.PathLike[str], Path], blocksize: int = 2**22) -> str:
+    lpath = stringify_path(lpath)
```
nicholasjng, line 33:
```suggestion
def md5_checksum(lpath: str | Path, blocksize: int = 2**22) -> str:
```

src/lakefs_spec/spec.py
```diff
@@ -203,7 +208,10 @@ def get_file(
 
         super().get_file(rpath=rpath, lpath=lpath, callback=callback, outfile=outfile, **kwargs)
 
-    def info(self, path: str, **kwargs: Any) -> dict[str, Any]:
+    def info(
+        self, path: Union[str, os.PathLike[str], pathlib.Path], **kwargs: Any
+    ) -> dict[str, Any]:
+        path_as_str = stringify_path(path)
```
nicholasjng, line 214: Any reason you're not doing `path = stringify_path(path)`? Is mypy / ruff unhappy? (That would save the changed lines below.)

nicholasjng (commented):
Please change the `{l,r}path` type hint to `str | os.PathLike[str]` everywhere (no Union). It is not necessary to include `pathlib.Path` in the type, since that has been supporting the `os.PathLike` interface since Python 3.6, and we are only supporting >=3.9.

(The reason fsspec does it in `stringify_path` is because they are supporting older Pythons than 3.6 as well.)

Other than that, together with the other two comments, all good!

src/lakefs_spec/spec.py
maxmynter: Yep. That was it. MyPy does not resolve that `path = stringify_path(path)` is now a string.

PR #188, opened 2023-11-30
Record test coverage on fsspec package [ENG-196]

AdrianoKF:
In order to gauge the effectiveness of our test suite, we should not
only measure coverage on our own code, but rather on the abstract
base classes for the file system and file implementations as well.

This commit adds the necessary parameters to the pytest invocation
in CI, which then ends up in the CodeCov reports.


AdrianoKF:
Merging this PR since the Codecov API token changed (due to the migration of the repo), hence it does not have any recorded history for the `main` branch.

We will perform any steps to improve coverage in a subsequent PR.

PR #190, opened 2023-11-30
Modify error translator to raise more informative errors

nicholasjng:
Prepares the `status`, `reason`, and `filename` arguments unconditionally for each observed API error.

This way, users who see translated errors actually know what the filename involved is instead of just getting the (less than ideal) lakeFS API reason.

One particular behavior is that of `fs.ls()` in case the branch/ref is not found: Here, unlike in the repository case, the branch is not part of the error message, which means just a generic "file not found" error.

We use OS Errnos for more exotic statuscodes, and HTTP codes for 401, 403 and 404 errors.

### DIFF
--- a/src/lakefs_spec/errors.py
+++ b/src/lakefs_spec/errors.py
@@ -39,15 +39,14 @@ def translate_lakefs_error(
 
     Parameters
     ----------
-
     error: lakefs_client.ApiException
         The exception returned by the lakeFS API.
-    rpath: str or None
+    rpath: str | None
         The remote resource path involved in the error.
-    message : str
+    message: str | None
         An error message to use for the returned exception.
          If not given, the error message returned by the lakeFS server is used instead.
-    set_cause : bool
+    set_cause: bool
         Whether to set the ``__cause__`` attribute to the previous exception if the exception is translated.
 
     Returns
@@ -61,12 +60,11 @@ def translate_lakefs_error(
     else:
         status, reason = error.code, error.reason
 
-    constructor = HTTP_CODE_TO_ERROR.get(status, partial(IOError, errno.EIO))
-
     emsg = f"{status} {reason}"
     if rpath:
         emsg += f": {rpath!r}"
 
+    constructor = HTTP_CODE_TO_ERROR.get(status, partial(IOError, errno.EIO))
     custom_exc = constructor(message or emsg)
 
     if set_cause:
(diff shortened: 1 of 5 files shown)


nicholasjng:
Repro script:

```python
def main():
    fs = LakeFSFileSystem()
    fs.get_file("quickstart/hello/hello.py", "hello.py")
```

On `main`, this results in:

```py
FileNotFoundError: HTTP404 (Not Found): not found
```

While on this PR branch:

```py
FileNotFoundError: [Errno 404] not found: 'quickstart/hello/hello.py'
```

nicholasjng:
Since the distinction between HTTP status codes and OS Errnos is completely arbitrary here: I'm also completely fine (and actually happier) with changing all error codes over to either type (HTTP codes vs. OS error codes), but they should be consistent.

nicholasjng:
What's also possible is to let both appear in the error message, e.g. like so:

```py
# errno 2 is ENOENT
FileNotFoundError: [Errno 2] 404 not found: 'quickstart/hello/hello.py'
```

Unfortunately I have not found a way to omit the actual errno bracket in the string representation if `reason` and `filename` are given.

src/lakefs_spec/errors.py
```diff
@@ -8,61 +8,67 @@
...
 
-HTTP_CODE_TO_ERROR: dict[int, type[OSError]] = {
+HTTP_CODE_TO_ERROR: dict[int, type[OSError] | partial[OSError]] = {
+    400: partial(IOError, errno.EINVAL),
     401: PermissionError,
     403: PermissionError,
     404: FileNotFoundError,
+    410: FileNotFoundError,  # Gone (temporarily / permanently unavailable)
+    416: partial(IOError, errno.EINVAL),  # invalid range
+    420: partial(IOError, errno.EBUSY),  # too many requests
```
AdrianoKF, line 24: We have to be a bit careful with the `errno` package since it might be platform-dependent. From the docs:

> Of the following list, symbols that are not used on the current platform are not defined by the module.

We should verify that these error numbers are in fact defined on all supported platforms (Linux/Darwin are easy enough, Windows I'm not so sure about)

nicholasjng: I went with the corresponding `s3fs` implementation, it seems none of the codes we use are affected: https://github.com/fsspec/s3fs/blob/main/s3fs/errors.py

maxmynter (approved):

PR #194, opened 2023-12-04
Drop `requirements.txt` from package lock files

nicholasjng:
It serves no purpose over `requirements-dev.txt`, which we advise installing for contributors, and is unused in package distribution.

Also clean up a stale pre-commit comment, and remove the `requirements.txt` compile hook.


nicholasjng:
Rebased, since the `requirements.txt` deletion ended up in the first (dev and docs update) commit.

PR #198, opened 2023-12-07
`ls`: Make non-recursive by default, correctly update dircache

AdrianoKF:
This PR combines various fixes to the handling of the directory listing cache and `ls` behavior in general:

## 🚨 Breaking Change: Make `ls` non-recursive by default
The `ls` method in its previous implementation would always return items in subdirectories recursively, which is not an intuitive behavior.

This PR introduces an optional `recursive` kwarg, which is set to `False` by default.

## `ls` dircache update
The `ls` command was not updating the dircache correctly, which could lead to incorrect results when listing the same directory multiple times.

In particular, the implementation would always add all results from the API call to the dircache for the directory, regardless of whether the results were already in the cache entry (essentially doubling the cache entry for each subsequent call).

This behavior would only be triggered by passing `refresh=True`, which bypasses the shortcut evaluation that returns a dircache entry directly.

## `rm` dircache invalidation
Additionally, this PR fixes the invalidation of the dircache as part of `rm`.

Removing a file from the repository obviously makes any cache directory listing of its parent directory invalid, which is now handled correctly (and tested).

### DIFF
--- a/src/lakefs_spec/client_helpers.py
+++ b/src/lakefs_spec/client_helpers.py
@@ -341,7 +341,7 @@ def reset_branch(client: LakeFSClient, repository: str, branch: str) -> None:
     repository: str
         Repository in which the specified branch is located.
     branch: str
-        Branch on which the commit should be reverted.
+        Branch to reset.
     """
     reset_creation = ResetCreation(type="reset")
     client.branches_api.reset_branch(
--- a/src/lakefs_spec/client_helpers.py
+++ b/src/lakefs_spec/client_helpers.py
@@ -21,6 +21,7 @@ from lakefs_sdk.models import (
     Ref,
     Repository,
     RepositoryCreation,
+    ResetCreation,
     RevertCreation,
     TagCreation,
 )
@@ -326,6 +327,25 @@ def revert(client: LakeFSClient, repository: str, branch: str, parent_number: in
     revert_creation = RevertCreation(ref=branch, parent_number=parent_number)
     client.branches_api.revert_branch(
         repository=repository, branch=branch, revert_creation=revert_creation
+    )
+
+
+def reset_branch(client: LakeFSClient, repository: str, branch: str) -> None:
+    """
+    Reset the specified branch to its head.
+
+    Parameters
+    ----------
+    client: LakeFSClient
+        lakeFS client object.
+    repository: str
+        Repository in which the specified branch is located.
+    branch: str
+        Branch on which the commit should be reverted.
+    """
+    reset_creation = ResetCreation(type="reset")
+    client.branches_api.reset_branch(
+        repository=repository, branch=branch, reset_creation=reset_creation
     )
 
 
--- a/src/lakefs_spec/spec.py
+++ b/src/lakefs_spec/spec.py
@@ -14,6 +14,7 @@ import os
 import urllib.error
 import urllib.request
 from contextlib import contextmanager
+from functools import partial
 from pathlib import Path
 from typing import Any, Generator, Iterable, Literal, cast
 
@@ -398,7 +399,7 @@ class LakeFSFileSystem(AbstractFileSystem):
             if cache_entry is not None:
                 if not detail:
                     return [e["name"] for e in cache_entry]
-                return cache_entry
+                return cache_entry[:]
 
         kwargs["prefix"] = prefix
 
@@ -422,14 +423,29 @@ class LakeFSFileSystem(AbstractFileSystem):
         if info:
             # assumes that the returned info is name-sorted.
             pp = self._parent(info[0]["name"])
+            info_copy = info[:]
             if pp in self.dircache:
                 # ls info has files not in cache, so we update them in the cache entry.
-                cache_entry = self.dircache[pp]
-                # extend the entry by the new ls results
-                cache_entry.extend(info)
+                cache_entry = self.dircache[pp][:]
+
+                # Overwrite existing entries in the cache with its updated values
+                for idx, entry in enumerate(cache_entry):
+
+                    def _matches(e, name):
+                        return e["name"] == name
+
+                    info_entry = next(
+                        filter(partial(_matches, name=entry["name"]), info_copy),
+                        None,
+                    )
+                    if not info_entry:
+                        continue
+                    cache_entry[idx] = info_entry
+                    info_copy.remove(info_entry)
+                cache_entry.extend(info_copy)
                 self.dircache[pp] = sorted(cache_entry, key=operator.itemgetter("name"))
             else:
-                self.dircache[pp] = info
+                self.dircache[pp] = info[:]
 
         if not detail:
             info = [o["name"] for o in info]
--- a/src/lakefs_spec/spec.py
+++ b/src/lakefs_spec/spec.py
@@ -446,6 +446,13 @@ class LakeFSFileSystem(AbstractFileSystem):
                     }
                 )
 
+        # Retry the API call with appended slash if the current result
+        # is just a single directory entry only (not its contents).
+        # This is useful to allow `ls("repo/branch/dir")` calls without
+        # a trailing slash.
+        if len(info) == 1 and info[0]["type"] == "directory":
+            return self.ls(path + "/", detail=detail, **kwargs | {"refresh": not use_dircache})
+
         # cache the info if not empty.
         if info:
             # assumes that the returned info is name-sorted.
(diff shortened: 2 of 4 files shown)


src/lakefs_spec/spec.py
```diff
@@ -422,14 +423,29 @@ def ls(self, path: str | os.PathLike[str], detail: bool = True, **kwargs: Any) -
...
+
+                    info_entry = next(
+                        filter(partial(_matches, name=entry["name"]), info_copy),
+                        None,
+                    )
+                    if not info_entry:
+                        continue
+                    cache_entry[idx] = info_entry
+                    info_copy.remove(info_entry)
+                cache_entry.extend(info_copy)
```
nicholasjng, line 445: How about something like a JavaScript index?

```py
def safe_index(l, elem):
    try:
        return l.index(elem)
    except IndexError:
        return -1

for info_entry in info_copy:
    idx = safe_index(cache_entry, info_entry)
    if idx >= 0:
        cache_entry[idx] = info_entry
        info_copy.remove(info_entry)

cache_entry.extend(info_copy)

Comment (AdrianoKF) [reply src/lakefs_spec/spec.py]: Completely rewrote that part of the code in the meantime, please have a look :)

File: src/lakefs_spec/client_helpers.py [hunk 2]
```diff
@@ -329,6 +330,25 @@ def revert(client: LakeFSClient, repository: str, branch: str, parent_number: in
...
+    Reset the specified branch to its head.
+
+    Parameters
+    ----------
+    client: LakeFSClient
+        lakeFS client object.
+    repository: str
+        Repository in which the specified branch is located.
+    branch: str
+        Branch on which the commit should be reverted.
```
nicholasjng, line 344: Copy-paste, please update

src/lakefs_spec/spec.py
```diff
@@ -348,17 +348,22 @@ def info(self, path: str | os.PathLike[str], **kwargs: Any) -> dict[str, Any]:
...
 
-        out = self.ls(path, detail=True, **kwargs)
+        out = self.ls(path, detail=True, recursive=True, **kwargs)
         if not out:
             raise FileNotFoundError(errno.ENOENT, os.strerror(errno.ENOENT), path)
 
         return {
             "name": path.rstrip("/"),
-            "size": sum(o.get("size", 0) for o in out),
+            "size": sum(o.get("size") or 0 for o in out),
```
nicholasjng, line 357: I prefer the old version - what was the problem with that? (This incurs another `None` to `int` comparison btw)

src/lakefs_spec/spec.py
```diff
@@ -398,14 +407,25 @@ def ls(self, path: str | os.PathLike[str], detail: bool = True, **kwargs: Any) -
             if cache_entry is not None:
                 if not detail:
                     return [e["name"] for e in cache_entry]
-                return cache_entry
+                return cache_entry[:]
+
+        recursive = kwargs.get("recursive", False)
+        if "recursive" in kwargs:
+            del kwargs["recursive"]
```
nicholasjng, line 414: You could use `recursive = kwargs.pop("recursive", False)` here.

src/lakefs_spec/spec.py
```diff
@@ -421,15 +441,37 @@ def ls(self, path: str | os.PathLike[str], detail: bool = True, **kwargs: Any) -
         # cache the info if not empty.
         if info:
             # assumes that the returned info is name-sorted.
+            # All directories included in this listing
+            directories = {str(Path(e["name"]).parent) for e in info}
+
             pp = self._parent(info[0]["name"])
```
nicholasjng, line 447: These two should be consistent to yield the same results, so either change to `self._parent(e["name"])` above or to the `Path().parent` approach below.

src/lakefs_spec/spec.py
```diff
@@ -646,6 +688,10 @@ def rm_file(self, path: str | os.PathLike[str]) -> None:
             self.client.objects_api.delete_object(
                 repository=repository, branch=branch, path=resource
             )
+            # Directory listing cache for the containing folder must be invalidated
+            parent_path = str(Path(path).parent)
+            if parent_path in self.dircache:
+                del self.dircache[parent_path]
```
nicholasjng, line 694: Also, this is `O(cachesize)` because of the `in` check, you could also `self.dircache.pop(parent_path, None)` for a (hopefully?) `O(1)` solution (not that it matters anywhere outside of lakeFS'ing in Google-sized monorepos)

src/lakefs_spec/spec.py
AdrianoKF: When not fetching recursively (i.e., `delimiter="/"`), the API returns `size=None`, which would lead to an attempt to call `__add__` with `None` and `int` params in the old version.

src/lakefs_spec/spec.py
AdrianoKF: Nice, didn't have that on my radar!

src/lakefs_spec/spec.py
AdrianoKF: Actually, `DirCache` implements `__contains__` as an element access in a `dict`, which has [average case complexity in `O(1)`](https://wiki.python.org/moin/TimeComplexity). However, as above, `pop` seems like the way to go here.

src/lakefs_spec/spec.py
AdrianoKF: Actually, the code line just above was a leftover from debugging; removed it.

src/lakefs_spec/spec.py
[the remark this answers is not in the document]
AdrianoKF: Changed it to `self._parent`, good catch.

src/lakefs_spec/client_helpers.py
AdrianoKF: Fixed in 7706b7b3751020aefdc38bc9a0f9cde251435573

src/lakefs_spec/spec.py
```diff
@@ -422,14 +440,33 @@ def ls(self, path: str | os.PathLike[str], detail: bool = True, **kwargs: Any) -
...
+
+                    cache_entry[old_idx] = new_entry
+                    info_copy.remove(new_entry)
+
+                # Add the remaining (new) entries to the cache
+                cache_entry.extend(info_copy)
                 self.dircache[pp] = sorted(cache_entry, key=operator.itemgetter("name"))
             else:
-                self.dircache[pp] = info
+                self.dircache[pp] = info[:]
```
nicholasjng, line 469: I'm unsure about the behavior here for subdirectories if `recursive=True`: On the one hand, they are contained in the parent, so their contents could go in the parent cache entry just fine.

On the other hand, they are also directories, so they could deserve their own cache entry.

We could just save the directory stub in the cache entry in the cache entry, and then materialize it by more cache lookups if `recursive=True`:

```py
data/a.txt
       /b.txt
       /dir/f.txt

-> fs.ls("data") = [{"name": "a.txt", ...}, {"name": "b.txt", ...}, {"name": "dir", ..., type="directory"}]
```
and then looking up `dir` in `fs.dircache` if `recursive=True`.

tests/test_ls.py
```diff
@@ -98,7 +100,7 @@ def test_ls_no_detail(fs: LakeFSFileSystem, repository: str) -> None:
 
     expected = [f"{resource}/lakes.source.md"]
     # first, verify the API fetch does the expected...
-    assert fs.ls(resource, detail=False) == expected
+    assert fs.ls(f"{resource}/", detail=False) == expected
```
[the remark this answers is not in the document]
AdrianoKF: Added the old behavior back in fc3dadb (which now automatically retries the API call in case it only returned a single directory entry). WDYT?

src/lakefs_spec/spec.py
AdrianoKF: I think this goes in the direction of my forgotten debugging statement in the original PR, which was collecting all the directories that were part of the `list_objects` response.

We currently have this, which I believe is not enough:

```python
            # assumes that the returned info is name-sorted.
            pp = self._parent(info[0]["name"])
```

Rather, we need to add every parent directory contained in the response to the cache, as you mentioned above. That's an easy enough fix, but it should be accompanied by a corresponding test.

nicholasjng (approved):

src/lakefs_spec/spec.py
AdrianoKF: Let's follow up separately on this one to unblock the merge of this PR.

PR #201, opened 2023-12-08
Improve test coverage on lakefs abstract base classes

AdrianoKF:
This PR adds a variety of smoke tests that aim to cover the default functionality provided by the abstract base classes from `fsspec`, namely `AbstractFileSystem` and `AbstractBufferedFile`.

In order to avoid the problem discovered by one of the tests and reported upstream ( https://github.com/treeverse/lakeFS/issues/7127), the version of lakeFS dependencies has been bumped to v1.3.1 in the CI and dev environments.

> [!NOTE]
> These tests are not meant to provide complete coverage of the base implementations, since these are tested as part of `fsspec` already.
> Rather, they aim to uncover any potential issues that arise from the mix of overwritten and default implementations and their interactions (e.g., quite a few implementations in the ABC rely on the return value from `ls`, which is overwritten in the library).

### DIFF
--- a/.github/workflows/python.yaml
+++ b/.github/workflows/python.yaml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     services:
       lakefs:
-        image: treeverse/lakefs:1.3.0
+        image: treeverse/lakefs:1.3.1
         ports:
           - 8000:8000
         env:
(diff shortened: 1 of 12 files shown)


maxmynter:
Do you mean just adding the `_strip_protocol` with `pass` as a method to our `LakeFSFileSystem`.
Or are other changes involded?

Because, if we just add the above, we get TypeErrors in `parse`, at least for `find` and resultingly methods that rely on find such as `glob`.  

Are there other implications to reinstating the overloads?

maxmynter:
As a user, I would expect for find to work also on a branch without providing a `ref` in the branch. Currently it does not work, as the trailing `/` gets stripped of and `parse` throws an error.

nicholasjng:
> I have spent some time with the `touch` `fsspec` command and did not get it to work. The problem is that the `objects_api.upload_object` in our `commit` function does not accept an empty content / mime type. If we set one, we get an error that the local file at the path is not found.
> 
> So I believe touch does not work with the `lakefs-sdk` API. I could also find nothing on these notes in the lakeFS docs.
> 
> Should we overwrite the method in our filesystem with a not implemented error? Or is there another way to make it work?

I raised this with the lakeFS devs in https://github.com/treeverse/lakeFS/issues/7127, which they resolved right away.

PR #203, opened 2023-12-11
Validate with docstrings with `pydoclint`

AdrianoKF:
This PR adds `pydoclint` as a dev dependency and pre-commit hook.

All non-confirming docstrings in the source code have been updated to pass the check.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,6 +49,7 @@ dev = [
     "pre-commit>=3.3.3",
     "pytest>=7.4.0",
     "pytest-cov>=4.1.0",
+    "pydoclint",
 ]
 docs = [
     "mkdocs",
@@ -123,3 +124,7 @@ exclude_dirs = ["tests", "docs/tutorials"]
 [tool.pytest.ini_options]
 log_cli = true
 log_cli_level = "WARNING"
+
+[tool.pydoclint]
+style = 'numpy'
+exclude = '\.git|venv'
--- a/src/lakefs_spec/spec.py
+++ b/src/lakefs_spec/spec.py
@@ -17,8 +17,9 @@ from contextlib import contextmanager
 from pathlib import Path
 from typing import Any, Generator, Iterable, Literal, cast
 
+import fsspec.callbacks
 from fsspec import filesystem
-from fsspec.callbacks import _DEFAULT_CALLBACK, Callback
+from fsspec.callbacks import _DEFAULT_CALLBACK
 from fsspec.spec import AbstractBufferedFile, AbstractFileSystem
 from fsspec.utils import stringify_path
 from lakefs_sdk import Configuration
@@ -163,7 +164,7 @@ class LakeFSFileSystem(AbstractFileSystem):
 
         Raises
         ------
-        ApiException
+        OSError
             Translated error from the lakeFS API call, if any.
         """
         try:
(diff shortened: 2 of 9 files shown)


src/lakefs_spec/spec.py
```diff
@@ -528,6 +548,11 @@ def _open(
         -------
         LakeFSFile
             A local file-like object ready to hold data to be received from / sent to a lakeFS server.
+
+        Raises
+        ------
+        NotImplementedError
+            If ``mode`` is not supported.
```
nicholasjng, line 555: I dislike this, because the `Literal` typing already clearly states which values are supported. Of course, it's technically what happens when you pass a bogus mode, but this is too verbose in my opinion.

pyproject.toml
```diff
@@ -49,6 +49,7 @@ dev = [
     "pre-commit>=3.3.3",
     "pytest>=7.4.0",
     "pytest-cov>=4.1.0",
+    "pydoclint",
```
nicholasjng, line 52: Why does it appear here? `pre-commit` is run isolated, and it does not seem to be used anywhere in the docs build / GH Actions.

src/lakefs_spec/spec.py
```diff
@@ -566,12 +591,17 @@ def put_file_to_blockstore(
...
             An fsspec callback to use during the operation. Can be used to report download progress.
         presign: bool
             Whether to use pre-signed URLs to upload the object via HTTP(S) using ``urllib.request``.
         storage_options: dict[str, Any] | None
             Additional file system configuration options to pass to the block storage file system.
+
+        Raises
+        ------
+        ValueError
+            If the blockstore type returned by the lakeFS API is not supported.
```
nicholasjng, line 604:
```suggestion
        ValueError
            If the blockstore type returned by the lakeFS API is not supported by fsspec.
```

src/lakefs_spec/spec.py
```diff
@@ -276,9 +291,9 @@ def get_file(
             The remote path to download to local storage. Must be a fully qualified lakeFS URI, and point to a single file.
         lpath: str | os.PathLike[str]
             The local path on disk to save the downloaded file to.
-        callback: fsspec.callbacks.Callback
+        callback: Callback
```
nicholasjng, line 294: Hmm, I prefer fully qualified types. Otherwise, someone reading the docstring might think that we define our own callback type somewhere.

pyproject.toml
AdrianoKF: IMO it's useful to just be able to run those tools in isolation (e.g., as part of an editor/IDE integration).

src/lakefs_spec/spec.py
AdrianoKF: I agree, but unfortunately, that cannot be configured in `pydoclint`.

One option would be to not use the imported `Callback` type in the actual signature, but rather its fully-qualified name (which `pydoclint` would then enforce in the docstring as well).

AdrianoKF: Did that in ebf7f2f71060e922c45fcc1ef2517a58466de53f

src/lakefs_spec/spec.py
AdrianoKF: It's an edge case (and I don't really care about it either way, TBH).

The code technically does raise this exception, so documenting it makes sense.
OTOH, the user never interacts with `_open` directly, so they won't even see this docstring in the first place - making the discussion around verbosity kinda moot.

In general, I like having an extra (automated) set of eyes on the docs, since the tool already caught a few places where the docstrings were not in sync with the code. In that sense, for me, the benefits outweigh the edge cases we're discussing here in the PR. (I have yet to see a user complain about documentation being too complete).

nicholasjng (approved):

PR #214, opened 2023-12-11
Add lakeFS server version guard to `LakeFSFilesystem.touch()`

### LINKED ISSUE #208: Gate functionality on lakeFS backend version
### Motivation

Follow-up of #201: We found out that `fs.touch()` will not work for any lakeFS server <1.3.1, since the backend had trouble with missing MIME types (see also https://github.com/treeverse/lakeFS/issues/7127).

We should raise an informative error in case any file system functionality is broken/unsupported for a backend version.

### Requested solution

Load the client's `VersionConfig` (https://pydocs-sdk.lakefs.io/docs/InternalApi.html#get_lake_fs_version), parse it out, check against a minimum version and maybe raise `NotImplementedError`s if the server version is too old.

Bonus: Cache the API response by host, so that multiple client requests can reuse the obtained value.

nicholasjng:
Guards against attempted `touch()` operations by checking the backend version against the minimum supported backend version. This is preferable to the rather cryptic `missing mime type` 500 error from the backend in lakeFS server versions before v1.3.1.

Also adds a test mocking the server config to return an unsupported version. Skipping on old lakeFS server versions is currently not implemented, but lower prio, since the CI runs on the latest Docker image.

Fixes #208.


src/lakefs_spec/spec.py
```diff
@@ -758,6 +758,38 @@ def rm_file(self, path: str | os.PathLike[str]) -> None:
...
+            Additional keyword arguments to pass to ``LakeFSFile.open()``.
+
+        Raises
+        ------
+        NotImplementedError
+            If the targeted lakeFS server version does not support `touch()` operations.
+        """
+        with self.wrapped_api_call():
+            version_string = self.client.config_api.get_config().version_config.version
+            version_tuple = tuple(int(t) for t in version_string.split("."))
```
AdrianoKF, line 781: Should we make this part of the constructor maybe and cache it? Having to guard code based on the server version might come up again in the future.

AdrianoKF:
> Skipping on old lakeFS server versions is currently not implemented, but lower prio, since the CI runs on the latest Docker image.

`@pytest.mark.xfail` can take a [boolean argument `condition`](https://docs.pytest.org/en/7.1.x/how-to/skipping.html#condition-parameter), which we could use to mark test in case of server versions <1.3.1.

src/lakefs_spec/spec.py
nicholasjng: Done in [71395c6](https://github.com/aai-institute/lakefs-spec/pull/214/commits/71395c6ba010c67da5a4021b252c4a97281e7b4e).

nicholasjng:
> > Skipping on old lakeFS server versions is currently not implemented, but lower prio, since the CI runs on the latest Docker image.
> 
> `@pytest.mark.xfail` can take a [boolean argument `condition`](https://docs.pytest.org/en/7.1.x/how-to/skipping.html#condition-parameter), which we could use to mark the test case as an expected failure in case of server versions <1.3.1.

Do you know of a way of skipping a test based on a fixture value? I was looking through the docs and found nothing. My idea is this:

```py
@pytest.skipif(lakefs_version < (1, 3, 1), reason="fs.touch() is not supported on old lakeFS server versions")
def test_touch():
    ...
```

AdrianoKF:
> > > Skipping on old lakeFS server versions is currently not implemented, but lower prio, since the CI runs on the latest Docker image.
> > 
> > 
> > `@pytest.mark.xfail` can take a [boolean argument `condition`](https://docs.pytest.org/en/7.1.x/how-to/skipping.html#condition-parameter), which we could use to mark the test case as an expected failure in case of server versions <1.3.1.
> 
> Do you know of a way of skipping a test based on a fixture value? I was looking through the docs and found nothing. My idea is this:
> 
> ```python
> @pytest.skipif(lakefs_version < (1, 3, 1), reason="fs.touch() is not supported on old lakeFS server versions")
> def test_touch():
>     ...
> ```

You could also call `pytest.xfail()` inside the test case, where you have access to the fixture value. I also couldn't find any other solution from a quick web search.

Issue #192, opened 2023-12-04
Working on docs for a longer time causes local failure of docs build because of too many requests on Weather API.


maxmynter:
Our recently introduced [change](https://github.com/aai-institute/lakefs-spec/pull/183) to handle the demo notebook with Jupytext causes a run of the entire notebook on every change to the docs via the hot reloading of `mkdocs serve`.

Since the note book is rebuild everytime from jupytext, it is executed on every hot reload. 
After some time of working on the docs, the notebook execution fails because the Open Meteo API blocks too many requests. Then one can not work on the docs.

Alternatives to be considered:
1) https://jupyter-cache.readthedocs.io/en/latest/  OR
2) implement client side caching for API access.

PR #220, opened 2023-12-13
Docs: Client-side caching for Open-Meteo API in tutorial notebook

AdrianoKF:
This PR adds a simple client-side caching function to the tutorial notebook, whichs prevents the same file from being downloaded multiple times.

This avoids hitting the API rate limit when working on the notebook locally.

Closes issue #192

### DIFF
--- a/docs/tutorials/demo_data_science_project.py
+++ b/docs/tutorials/demo_data_science_project.py
@@ -34,7 +34,7 @@ We will do the following:
 
 !!! tip "Local Execution"
     If you want to execute the code in this tutorial as a Jupyter notebook yourself, download the `demo_data_science_project.py` file from the lakeFS-spec repository.
-    
+
     You can then convert the Python file to a notebook using [Jupytext](https://jupytext.readthedocs.io/en/latest/using-cli.html) using the following command: `jupytext --to notebook demo_data_science_project.py`.
 
 This tutorial assumes that you have installed lakeFS-spec in a virtual environment, and that you have followed the [quickstart guide](../quickstart.md) to set up a local lakeFS instance.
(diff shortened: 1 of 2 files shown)


docs/tutorials/demo_data_science_project.py
```diff
@@ -109,15 +111,28 @@
...
 
+
 # %%
-outfile, _ = urllib.request.urlretrieve(
-    "https://archive-api.open-meteo.com/v1/archive?latitude=52.52&longitude=13.41&start_date=2010-01-01&end_date=2010-12-31&hourly=temperature_2m,relativehumidity_2m,rain,pressure_msl,surface_pressure,cloudcover,cloudcover_low,cloudcover_mid,cloudcover_high,windspeed_10m,windspeed_100m,winddirection_10m,winddirection_100m"
-)
+def _maybe_urlretrieve(url: str, filename: str) -> str:
+    # Avoid API rate limit errors by downloading to a fixed local location
+    destination = Path(tempfile.gettempdir()) / "lakefs-spec-tutorials" / filename
+    destination.parent.mkdir(exist_ok=True, parents=True)
```
nicholasjng, line 124: What's the lifecycle of that `tempdir`? Is there any manual cleanup necessary?

AdrianoKF: Up to the host OS. In fact, cleanup is the opposite of what we want, in order to prevent the API from being accessed multiple times when rebuilding the notebook (which happens in different processes).

I know, it's kinda hacky, but if you run of diskspace on your `$TMPDIR`, you should blame your OS for not cleaning it up on reboot/schedule 😅

nicholasjng: But this gives the same directory on multiple interpreter sessions, i.e. it caches on a rebuild? (I don't quite understand how, which is why I'm asking)

nicholasjng (approved):

PR #221, opened 2023-12-13
Add overloads for `LakeFSFileSystem.ls()`

nicholasjng:
Specializes the return type as `list[dict[str, Any]]` for `detail=True`, and `list[str]` for `detail=False`.

Unfortunately, since the base class has a keyword argument default, the `Literal[False]` overload is still marked by the IDE.

Furthermore, the typing is changed to `list[str] | list[dict[str, Any]]`, since that is only equivalent to `list[str | dict[str, Any]]` in case of a homogeneous list. This is our case, but the former list union type explicitly conveys type homogeneity.

### DIFF
--- a/src/lakefs_spec/spec.py
+++ b/src/lakefs_spec/spec.py
@@ -406,12 +406,39 @@ class LakeFSFileSystem(AbstractFileSystem):
             "type": "directory",
         }
 
+    @overload
+    def ls(
+        self,
+        path: str | os.PathLike[str],
+        detail: Literal[True] = ...,
+        **kwargs: Any,
+    ) -> list[dict[str, Any]]:
+        ...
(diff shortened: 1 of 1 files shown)


nicholasjng:
Repro:

```py
from typing import reveal_type

def main():

    fs = LakeFSFileSystem()
    resource = "lakefs-spec-tests/main/data/"
    res = fs.ls(resource)

    res2 = fs.ls(resource, detail=False)

    reveal_type(res)

if __name__ == "__main__":
    main()
```

On the PR branch, `res2` is hinted as `list[str]`, which is nicer than what we had before. Unfortunately, due to the absence of these hints on the base class, my Pycharm still complains on overload 2 due to no default argument.

PR #222, opened 2023-12-13
Remove dunder methods from API docs

nicholasjng:
Since `LakeFSFile.__del__` was overloaded, it appeared in the docs, as mkdocstrings only removes single-underscore methods from the rendered documentation.

`__init__` methods are excluded with a different flag already, so to fix, just extend the filter regex to double-leading-underscore methods.


AdrianoKF (approved):
LGTM

PR #224, opened 2023-12-14
Correctly handle dircache for recursive `ls` calls

### LINKED ISSUE #200: Improve `ls` dircache handling for recursive listings
The updating of dircache entries in `ls` when dealing with recursive listings (which might contain files across multiple parent directories) needs to be improved.

The current implementation does not add all subdirectories from the recursive listing to the cache individually, but rather just the first one encountered.

---
Original discussion:

> I think this goes in the direction of my forgotten debugging statement in the original PR, which was collecting all the directories that were part of the `list_objects` response.

We currently have [this](https://github.com/aai-institute/lakefs-spec/blob/c31b405eda8cd5a0d1f8759181804a010f9781ac/src/lakefs_spec/spec.py#L462-L463), which I believe is not enough:

```python
# assumes that the returned info is name-sorted.
pp = self._parent(info[0]["name"])
```

Rather, we need to add every parent directory contained in the response to the cache (not just that of the first item), as you mentioned above. That's an easy enough fix, but it should be accompanied by a corresponding test.

_Originally posted by @AdrianoKF in https://github.com/aai-institute/lakefs-spec/pull/198#discussion_r1420246607_

AdrianoKF:
The previous implementation of `ls` would only update the dircache for a single directory, based on the order of info entries returned from the API.

The new implementation correctly adds all directory entries to the dircache, and makes sure that the responses returned from the cache correctly honor the `recursive` flag.

Closes #200

### DIFF
--- a/src/lakefs_spec/spec.py
+++ b/src/lakefs_spec/spec.py
@@ -443,7 +443,7 @@ class LakeFSFileSystem(AbstractFileSystem):
             cache_entry.extend(dir_info)
             self.dircache[pp] = sorted(cache_entry, key=operator.itemgetter("name"))
 
-    def _ls_from_cache(self, path: str, recursive: bool = False) -> Any | list | None:
+    def _ls_from_cache(self, path: str, recursive: bool = False) -> list[dict[str, Any]] | None:
         """Override of ``AbstractFileSystem._ls_from_cache`` with support for recursive listings."""
         if not recursive:
             return super()._ls_from_cache(path)
@@ -455,7 +455,9 @@ class LakeFSFileSystem(AbstractFileSystem):
             if result is None:
                 result = []
             result.extend(files)
-        return result
+        if not result:
+            return result
+        return sorted(result, key=operator.itemgetter("name"))
 
     @overload
     def ls(
@@ -528,10 +530,7 @@ class LakeFSFileSystem(AbstractFileSystem):
         recursive = kwargs.pop("recursive", False)
 
         # Try lookup in dircache unless explicitly disabled by `refresh=True` kwarg
-        use_dircache = True
-        if "refresh" in kwargs:
-            use_dircache = not kwargs["refresh"]
-            del kwargs["refresh"]  # cannot be forwarded to the API
+        use_dircache = not kwargs.pop("refresh", False)
 
         if use_dircache:
             cache_entry: list[Any] | None = None
@@ -584,9 +583,11 @@ class LakeFSFileSystem(AbstractFileSystem):
             )
 
         if recursive:
-            subdirs = {
-                self._parent(o["name"]) for o in info if self._parent(o["name"]) != path.rstrip("/")
-            }
+            # To make recursive ls behave identical to the non-recursive case,
+            # add back virtual `directory` entries, which are only returned by
+            # the lakeFS API when querying non-recursively.
+            here = self._strip_protocol(path).rstrip("/")
+            subdirs = {parent for o in info if (parent := self._parent(o["name"])) != here}
             for subdir in subdirs:
                 info.append(
                     {
--- a/src/lakefs_spec/spec.py
+++ b/src/lakefs_spec/spec.py
@@ -411,8 +411,7 @@ class LakeFSFileSystem(AbstractFileSystem):
         parents = {self._parent(i["name"]) for i in info}
         for pp in parents:
             # subset of info entries which are direct descendants of `parent`
-            dir_info = [i for i in info if self._parent(i["name"]) == pp]
-
+            dir_info = [i for i in info if self._parent(i["name"].rstrip("/")) == pp]
             if pp not in self.dircache:
                 self.dircache[pp] = dir_info
                 continue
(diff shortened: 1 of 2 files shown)


tests/test_ls.py
```diff
@@ -147,12 +145,57 @@ def test_ls_dircache_remove_cached(
...
+
+def test_ls_dircache_recursive(
+    fs: LakeFSFileSystem,
+    repository: str,
+    temp_branch: str,
+) -> None:
+    prefix = f"{repository}/{temp_branch}"
+
+    # (1) Basic recursive dircache filling
+    fs.dircache.clear()
```
nicholasjng, line 164: Should we move this to the `fs` fixture?

src/lakefs_spec/spec.py
```diff
@@ -406,6 +406,57 @@ def info(self, path: str | os.PathLike[str], **kwargs: Any) -> dict[str, Any]:
...
+                new_entry = next(e for e in info if e["name"] == name)
+
+                cache_entry[old_idx] = new_entry
+                dir_info.remove(new_entry)
+
+            # Add the remaining (new) entries to the cache
+            cache_entry.extend(dir_info)
+            self.dircache[pp] = sorted(cache_entry, key=operator.itemgetter("name"))
+
+    def _ls_from_cache(self, path: str, recursive: bool = False) -> Any | list | None:
```
nicholasjng, line 446: Return type should be `list[dict[str, Any]] | None`.

src/lakefs_spec/spec.py
```diff
@@ -406,6 +406,57 @@ def info(self, path: str | os.PathLike[str], **kwargs: Any) -> dict[str, Any]:
...
+        """Override of ``AbstractFileSystem._ls_from_cache`` with support for recursive listings."""
+        if not recursive:
+            return super()._ls_from_cache(path)
+
+        result = None
+        for key, files in self.dircache.items():
+            if not (key.startswith(path) or path == key + "/"):
+                continue
```
nicholasjng, line 454: Are we consistent with keeping trailing slashes in the cache? Then this can be simplified to the latter condition.

tests/test_ls.py
```diff
@@ -147,12 +145,57 @@ def test_ls_dircache_remove_cached(
...
+        assert all([fs._parent(v["name"].rstrip("/")) == cache_dir for v in files])
+
+    # (2) Dircache correctness, recursive
+    cached_listing_recursive = fs.ls(prefix + "/", recursive=True)
+    # Recursive listing from cache must contain all items, ordering need not be preserved
+    assert {o["name"] for o in cached_listing_recursive} == {o["name"] for o in listing}
+
+    # (3) Dircache correctness, non-recursive
+    cached_listing_nonrecursive = fs.ls(prefix + "/", recursive=False)
+    # Non-recursive listing from cache most only contain direct descendants of the listed directory
```
nicholasjng, line 179:
```suggestion
    # Non-recursive listing from cache must only contain direct descendants of the listed directory
```

src/lakefs_spec/spec.py
```diff
@@ -406,6 +406,57 @@ def info(self, path: str | os.PathLike[str], **kwargs: Any) -> dict[str, Any]:
...
+            if result is None:
+                result = []
+            result.extend(files)
+        return result
```
nicholasjng, line 458: Results should be sorted again here, because the keys will not be in alphabetic order.

src/lakefs_spec/spec.py
```diff
@@ -474,6 +525,8 @@ def _api_path_type_to_info(path_type: str) -> Literal["file", "directory"]:
         path = cast(str, stringify_path(path))
         repository, ref, prefix = parse(path)
 
+        recursive = kwargs.pop("recursive", False)
+
         # Try lookup in dircache unless explicitly disabled by `refresh=True` kwarg
         use_dircache = True
```
nicholasjng, line 531:
```suggestion
        use_dircache = not kwargs.pop("refresh", False)
```

then L532-534 can also go.

tests/test_ls.py
```diff
@@ -124,19 +124,17 @@ def test_ls_dircache_remove_uncached(
...
 
-        # ... and through the cache (which should have been updated above)
-        listing_post = fs.ls(resource)
-        assert len(listing_post) == len(listing_pre) - 1
-    finally:
-        client_helpers.reset_branch(fs.client, repository, temp_branch)
+    # ... and through the cache (which should have been updated above)
+    listing_post = fs.ls(resource)
+    assert len(listing_post) == len(listing_pre) - 1
+    client_helpers.reset_branch(fs.client, repository, temp_branch)
```
nicholasjng, line 137: I'm guessing this is also inconsequential like in the other test below, where you deleted the branch reset.

nicholasjng (commented):
Nice work!

src/lakefs_spec/spec.py
AdrianoKF: Addressed in 926fb3be2d1eee23b75dfc9514e8570acff35519

src/lakefs_spec/spec.py
AdrianoKF: Addressed in 926fb3be2d1eee23b75dfc9514e8570acff35519

tests/test_ls.py
AdrianoKF: Addressed in 926fb3be2d1eee23b75dfc9514e8570acff35519

tests/test_ls.py
AdrianoKF: It's not even required, since we clear the instance cache and return a fresh FS instance in the fixture.

Removed in 926fb3be2d1eee23b75dfc9514e8570acff35519.

tests/test_ls.py
AdrianoKF: Removed in 926fb3be2d1eee23b75dfc9514e8570acff35519.

src/lakefs_spec/spec.py
AdrianoKF: Addressed in 926fb3be2d1eee23b75dfc9514e8570acff35519.

src/lakefs_spec/spec.py
AdrianoKF: I tried removing it, but it does not seem like it. There's potentially a difference when it comes to the repository root itself.

nicholasjng (approved):

Issue #207, opened 2023-12-11
Investigate ways of simplifying transaction `Placeholder` class


Maciej818:
### Motivation

Simplify working with the `Placeholder` objects used by the transaction feature.

### Requested solution

We should investigate how to simplify working with the `Placeholder` objects used by the transaction feature (e.g., removing the need to call `placeholder.unwrap()` to obtain the underlying object).

### Proposed implementation approach

@AdrianoKF has a quick prototype based on the `wrapt` package, which might serve as a starting point:
```
from dataclasses import dataclass
from typing import Generic, TypeVar

import wrapt

T = TypeVar("T")


@dataclass
class Data:
    message: str


class Future(Generic[T], Data, wrapt.ObjectProxy):
    @property
    def is_filled(self) -> bool:
        return self.__wrapped__ is not None


a = Data("hello")
wrapped = Future(a)
print(a, wrapped)
print(a.message, wrapped.message)
print(wrapped.is_filled)
print(a is wrapped)

print("-" * 80)

empty_wrapper = Future[Data](None)
print(empty_wrapper.is_filled)
``` 

Resources:

[wrapt — wrapt 1.16.0 documentation](https://wrapt.readthedocs.io/en/master/index.html)

### Acceptance Criteria (mandatory to start implementation)

- [ ] Simplify tx `Placeholder` class.

### Definition of Done (mandatory to close an issue as completed)

- [ ] All applicable Definition of Done criteria met

PR #225, opened 2023-12-14
Simplify transaction `Placeholder` class

maxmynter:
## Main Changes

Using `wrapt` to refactor the `Placeholder` class with `wrapt.ObjectProxy` as a transparent proxy. 

Remove the need for `unwrap` via overriding of `__getattr__`.

Delete `unwrap_placeholders` utility function.

Make the `available` method an attribute as it needs no computation input.

## Other Changes
 Update maintainer e-mails to institute addresses.

Closes https://github.com/aai-institute/lakefs-spec/issues/207

### DIFF
--- a/src/lakefs_spec/transaction.py
+++ b/src/lakefs_spec/transaction.py
@@ -49,6 +49,16 @@ class Placeholder(Generic[T], wrapt.ObjectProxy):
             )
         return getattr(self.__wrapped__, name)
 
+    def __str__(self):
+        """Override to return the string representation of the wrapped object."""
+        if self.__wrapped__ is None:
+            raise RuntimeError(f"Placeholder of '{type(self).__name__}' is unfilled.")
+        return str(self.__wrapped__)
+
+    def __repr__(self):
+        """Override representation to represent the wrapped object."""
+        return repr(self.__wrapped__)
+
 
 class LakeFSTransaction(Transaction):
     """
--- a/src/lakefs_spec/transaction.py
+++ b/src/lakefs_spec/transaction.py
@@ -26,40 +26,26 @@ if TYPE_CHECKING:
 
 
 @dataclass
-class Placeholder(wrapt.ObjectProxy, Generic[T]):
+class Placeholder(Generic[T], wrapt.ObjectProxy):
     """A generic placeholder for a value computed by the lakeFS server in a versioning operation during a transaction."""
 
     def __init__(self, wrapped: T | None = None):
-        super(Placeholder, self).__init__(wrapped)
+        super().__init__(wrapped)
+        self._self_value = wrapped
 
     @property
     def available(self) -> bool:
         """Whether the wrapped value is available, i.e. already computed."""
         return self.__wrapped__ is not None
 
-    def set_value(self, value: T) -> None:
+    @property
+    def value(self):
+        return self.__wrapped__
+
+    @value.setter
+    def value(self, val: T) -> None:
         """Fill in the placeholder. Not meant to be called directly except in the completion of the transaction."""
-        self.__wrapped__ = value
-
-    def __getattr__(self, name):
-        """Override __getattr__ to access the attributes of the wrapped object directly."""
-        if self.__wrapped__ is None:
-            raise RuntimeError(
-                f" '{name}' attribute of object '{type(self).__name__}' is unfilled."
-            )
-        return self.__wrapped__.__getattribute__(name)
-
-    def __str__(self):
-        """Override to return the string representation of the wrapped object."""
-        if self.__wrapped__ is None:
-            return "<Placeholder>"
-        return self.__wrapped__.__str__
-
-    def __repr__(self):
-        """Override representation to represent the wrapped object."""
-        if self.__wrapped__ is None:
-            return "<Placeholder>"
-        return self.__wrapped__.__repr__
+        self.__wrapped__ = val
 
 
 class LakeFSTransaction(Transaction):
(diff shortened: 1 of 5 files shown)


nicholasjng (commented):
Two questions: a) What are the typing implications for this? (I remember reading something in the wrapt docs about this, I guess I cannot do `isinstance(sha, Ref)` on a placeholder?)

And b) I would like to be able to do something like `rpath=f"{repo}/{sha}/file.txt` with a sha computed via transaction - does this work yet? (We might have to override the `__repr__` or `__str__` methods to just call those of the wrapped object)

src/lakefs_spec/transaction.py
```diff
@@ -25,30 +26,28 @@
...
 
 @dataclass
-class Placeholder(Generic[T]):
+class Placeholder(Generic[T], wrapt.ObjectProxy):
     """A generic placeholder for a value computed by the lakeFS server in a versioning operation during a transaction."""
 
-    value: T | None = None
-    """The abstract value. Set only on completion of the versioning operation during the defining transaction."""
+    def __init__(self, wrapped: T = None):
+        super(Placeholder, self).__init__(wrapped)
```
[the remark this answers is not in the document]
maxmynter: The type will be returned as the type upon instantiation. In our case that would be `None`  before the placeholder is filled. I.e. `isinstance(sha, Ref)` is `False` and `isinstance(sha, None)` is `True` pre-instantiation. 

[Link to Docs Section](https://wrapt.readthedocs.io/en/master/wrappers.html#type-comparison)

maxmynter:
Working with string and representations should work now. I've added the `expected` argument that enables a more descriptive logging of what the placeholder is placeholding if `str` or `repr` is called. 

I did not get overriding `__instancecheck__` to work as the wrapt object takes precedence. It should somehow be possible. So lets discuss if investing additional time is worth it (plus maybe some pairing). 

For now `isinstance(sha, Ref)` on a placeholder gives False.

src/lakefs_spec/transaction.py
```diff
@@ -25,30 +26,43 @@
...
-class Placeholder(Generic[T]):
+class Placeholder(wrapt.ObjectProxy, Generic[T]):
     """A generic placeholder for a value computed by the lakeFS server in a versioning operation during a transaction."""
 
-    value: T | None = None
-    """The abstract value. Set only on completion of the versioning operation during the defining transaction."""
+    _expect: Type[T] | None = None
 
-    def available(self):
+    def __init__(self, wrapped: T | None = None, expect: Type[T] | None = None):
```
nicholasjng, line 34: I'd revert this - it's a nice idea, but not terribly useful, and it's extra work to maintain. If users run into trouble, they can read the place-held type from the transaction helper type hint / docstring.

src/lakefs_spec/transaction.py
```diff
@@ -25,30 +26,43 @@
...
-        if self.value is None:
-            raise RuntimeError("placeholder unfilled")
-        return self.value
+    def __getattr__(self, name):
+        """Override __getattr__ to access the attributes of the wrapped object directly."""
+        if self.__wrapped__ is None:
+            raise RuntimeError(
+                f" '{name}' attribute of object '{type(self).__name__}' is unfilled."
+            )
+        return getattr(self.__wrapped__, name)
```
nicholasjng, line 53: I'd prefer all of these to be class method calls instead of builtins, i.e. `self.__wrapped__.__getattr__(name)`, and similarly below with `__repr__` and `__str__`.

tests/test_transactions.py
```diff
@@ -166,3 +169,37 @@ def test_transaction_failure(
...
+    random_file = random_file_factory.make()
+
+    lpath = str(random_file)
+    rpath = f"{repository}/{temp_branch}/{random_file.name}"
+
+    message = f"Add file {random_file.name}"
+
+    with fs.transaction as tx:
+        fs.put_file(lpath, rpath)
+        sha = tx.commit(repository, temp_branch, message=message)
```
nicholasjng, line 189: Woah, that just hit me - `sha` will be `None` here in case of no unstaged changes, right? That's a bug.

nicholasjng (commented):
Great work, you're almost there!

tests/test_transactions.py
maxmynter: What is the desired behaviour? Do we want it to be a `Placeholder` instance or `Commit` (or whatever we are placeholding)?

src/lakefs_spec/transaction.py
maxmynter: Sure. 

Why your preference of methods over builtins?

src/lakefs_spec/transaction.py
maxmynter: Addressed in [848d28d](https://github.com/aai-institute/lakefs-spec/pull/225/commits/848d28df5dff14137351a46261cb4ba323713a74)

src/lakefs_spec/transaction.py
maxmynter: Adressed in [cebc153](https://github.com/aai-institute/lakefs-spec/pull/225/commits/cebc153db0f3087e5a87d2d9382f1c1b4dcd80b8)

nicholasjng (approved):
Nice!

tests/test_transactions.py
nicholasjng: Actually, `client_helpers.commit` will return the tip of the branch if no unstaged changes exist, so this is not a bug.

Issue #195, opened 2023-12-06
Integrate new `lakefs` SDK wrapper into the project


nicholasjng:
The new `lakefs` SDK wrapper proposal is available in beta: https://pypi.org/project/lakefs/

Should therefore be installable via `pip install --pre lakefs`.

Even though it is not completely ready yet, we can see if there is functionality that can already supersede some of our APIs (I'm thinking the client helpers especially), and save our learnings in this thread to ease migration to the `lakefs` package later.

nicholasjng:
One particularly interesting thing will be how to handle client autodiscovery: The wrapper APIs support omission of the client everywhere, which prompts an autodiscovery mechanism similarly to what we do now, but as a stateful session opened via `init`.

Will we use the "one file system, one session" approach or should we opt to reuse a session?

AdrianoKF:
Update: the first stable releases, v0.1.0 and v0.1.1, have just landed on PyPI: https://pypi.org/project/lakefs/#history

nicholasjng:
Closed by #231.

PR #231, opened 2023-12-19
Migrate the package to use the new `lakefs` SDK wrapper library

nicholasjng:
Addresses #195.

Progress:

- [x] Migrated `LakeFSFileSystem`
- [x] Migrated `LakeFSFile`
- [x] Deleted `lakefs_spec.config`
- [x] Migrated `conftest`
- [x] Migrated unit tests
- [x] Migrated smoke tests
- [x] Deleted `lakefs_spec.client_helpers`
- [x] Migrated `LakeFSTransaction`

A point of contention is the client initialization. ~I removed the client configuration attributes for now, but they could be added back depending on whether we want to still support them.~
EDIT: The attributes are added back, but the configfile is pending a decision, as stated below.

As for the config file - `lakefs` currently only supports the default location `$HOME/.lakectl.yaml`, so if we take over their behavior, the `configfile` argument is not useful anymore until they allow other files.


nicholasjng:
Hahaha, this change deletes so many lines that the new coverage falls below the current total coverage target!

src/lakefs_spec/spec.py
```diff
@@ -88,42 +84,43 @@ def __init__(
...
-            host=host or os.getenv("LAKEFS_HOST") or lakectl_config.host,
-            api_key=api_key or os.getenv("LAKEFS_API_KEY"),
-            api_key_prefix=api_key_prefix or os.getenv("LAKEFS_API_KEY_PREFIX"),
-            access_token=access_token or os.getenv("LAKEFS_ACCESS_TOKEN"),
-            username=username or os.getenv("LAKEFS_USERNAME") or lakectl_config.username,
-            password=password or os.getenv("LAKEFS_PASSWORD") or lakectl_config.password,
-            ssl_ca_cert=ssl_ca_cert or os.getenv("LAKEFS_SSL_CA_CERT"),
-        )
+            # TODO: `lakefs` does not consider configfiles other than ~/.lakectl.yaml,
+            #  do we keep any file or do we drop configfile support and defer?
```
maxmynter, line 101: I don't see where other configurations would be necessary. So I'd be in favour of deferment

maxmynter (approved):
Nice!

src/lakefs_spec/spec.py
nicholasjng: Removed `lakefs_spec.config` in [71bd77d](https://github.com/aai-institute/lakefs-spec/pull/231/commits/71bd77d2717372b1f4f9ff296aa7ab9dbeff9362).

Issue #233, opened 2023-12-20
Follow-ups of the SDK wrapper change


nicholasjng:
A few things were left over from the SDK wrapper migration in #231.

- [x] Implementing file reads and writes with lakefs objects, see https://github.com/treeverse/lakeFS/blob/master/clients/python-wrapper/lakefs/object.py for details. (#240)
- [x] Implementing deletes via `lakefs.Branch(...).delete_objects(*rpaths)`, both `rm_file` and `rm`. See https://github.com/treeverse/lakeFS/blob/af2394abd136a1c33b7dd2d3cff29d0fde40c822/clients/python-wrapper/lakefs/branch.py#L174 and the usage example in the docstring.
- [x] Implementing object copies as writes between branches if there's a nice API on the `Branch` class, otherwise leave as is. (#240)
- [ ] Find a nicer implementation of `rev_parse` for the demo than the one I made.
- [x] Implement a generic type `ConcreteOrAbstract = T | Placeholder[T]` (name TBD) and use it to improve `ref` typing on some of the transaction APIs. (#245)
- [x] Wrap `LakeFSFile.details` in the `wrapped_api_call` context instead of the super call in `(get|put)_file`. (#240)

We should also control all leftover `lakefs_sdk` imports to check if an equivalent name or better solution exists in the `lakefs` package, and if so, use that instead.

nicholasjng:
That being said, do we even need a generic abstract value with `wrapt` if `ininstance` checks hold anyway for placeholders? That should relax the interface typing, so point 5) might not be necessary.

nicholasjng:
I raised the remaining point 4 (better revision parsing) with treeverse in https://github.com/treeverse/lakeFS/issues/7232 - I consider this the appropriate place for the implementation (if it is happening), but we can roll our own anytime if they decide against it.

PR #236, opened 2023-12-21
Implement deletes using the lakeFS branch API

maxmynter:
Addresses point two of #233.


nicholasjng:
Note for posteriority: `rm_file` is now a shim to `rm`, which uses a lakeFS branch API designed for bulk deletions. We can thus skip covering it explicitly in tests.

PR #237, opened 2023-12-21
Sdk-follow-up

maxmynter:
Adresses point 1 from #233 

Currently blocked by the fact that the `lakefs` API does not provide a way to discard an opened file.


nicholasjng:
Re: our discussion yesterday, I opened https://github.com/treeverse/lakeFS/issues/7195 as a request for discard APIs on both reader and writer.

For now, the read/write attempt is blocked by some sharp edges in the `lakefs` file-like interfaces, so I think it's best to leave that point as last and focus on the remainder instead.

nicholasjng:
Superseded by #240, which contains two commits cherrypicked from here.

PR #238, opened 2023-12-22
Add custom lakeFS-spec logo/favicon

AdrianoKF:
The PR adds a custom lakeFS-spec logo and favicon to the project and docs:

- GitHub and PyPI README with dark/light mode support
- Docs: Redesigned landing page (no more headline, added logo) and favicon
- Raw Inkscape SVG under `assets/` (note that this is a [Multipage SVG](https://wiki.inkscape.org/wiki/Multipage), everything but the first page is only visible in Inkscape)

The dark/light mode switch using the HTML `<picture>` element with a media query should be compatible with PyPI's description renderer, based on my short experimentation.

As for the usage, I would suggest to use the light-mode version (i.e., the one with the darker text color) as the official logo, unless we specifically need to place it against a dark background.

## GitHub Readme

image
image

## Docs Landing page

image
image


nicholasjng:
Awesome stuff! But the badges should go underneath the logo in my opinion.

AdrianoKF:
> Awesome stuff! But the badges should go underneath the logo in my opinion.

Done in 70733f7 (see https://github.com/aai-institute/lakefs-spec/blob/70733f7f3bf8406ecc7f25ade63a27b7e9f7553b/README.md).

nicholasjng:
Could we move it under the headline (lakeFS-spec: An fsspec backend for lakeFS) too? I think that's standard practice.

See e.g. https://github.com/google/jax/#jax-autograd-and-xla or https://github.com/ml-explore/mlx

nicholasjng (approved):

AdrianoKF:
> Could we move it under the headline (lakeFS-spec: An fsspec backend for lakeFS) too? I think that's standard practice.
> 
> See e.g. https://github.com/google/jax/#jax-autograd-and-xla or https://github.com/ml-explore/mlx

See 9c25cd8 / https://github.com/aai-institute/lakefs-spec/blob/9c25cd83956ea7cb235da8a147479db4baa78a05/README.md. Acually, this looks much cleaner, thanks for the feedback!

PR #247, opened 2024-01-04
Update bug-report.yaml

Maciej818:
In the bug template there are lakeFS version requested twice (which results in error as labels are not unique). One 'lakeFS' version request to be replaced with requesting 'lakeFS-sdk' version.


Maciej818:
As agreed with @janwillemkl-aai: deleted lakeFS-sdk version as lakeFS version is sufficient.

PR #250, opened 2024-01-10
Replace `fs._open()` with proper `fs.open()` implementation

nicholasjng:
This is sensible because a) the builtin `fs.open()` was giving wrong type hints of `AbstractBufferedFile`s, which we no longer use, and b) the new lakeFS APIs allow native handling of non-bytemode open(), which the base `fs.open()` needs a TextIOWrapper for.

No new tests, since all file-system APIs are already being tested. What is missing are integration tests to pandas, polars, etc., since those all rely on `fs.open()`.

### DIFF
--- a/src/lakefs_spec/spec.py
+++ b/src/lakefs_spec/spec.py
@@ -572,29 +572,35 @@ class LakeFSFileSystem(AbstractFileSystem):
 
         return info
 
-    def _open(
+    def open(
         self,
         path: str | os.PathLike[str],
-        mode: Literal["rb", "wb", "xb"] = "rb",
+        mode: Literal["r", "rb", "rt", "w", "wb", "wt", "x", "xb", "xt"] = "rb",
+        pre_sign: bool = False,
+        content_type: str | None = None,
+        metadata: dict[str, str] | None = None,
+        autocommit: bool = True,
         **kwargs: Any,
     ) -> LakeFSIOBase:
         """
-        Dispatch a lakeFS file (local buffer on disk) for the given remote path for up- or downloads depending on ``mode``.
-
-        Internal only, called by ``AbstractFileSystem.open()``.
+        Dispatch a lakeFS file-like object (local buffer on disk) for the given remote path for up- or downloads depending on ``mode``.
 
         Parameters
         ----------
         path: str | os.PathLike[str]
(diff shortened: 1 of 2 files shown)


nicholasjng:
For reviewers:

1) Should we explicitly keep allowing `{r,w,x}t` modes if the only thing we do is silently coerce them into `{r,w,x}b`? We have one such example in the readme code, which would need to be updated.
2) What's your feeling on finally adding an integration test for the data science zoo (see the commit message)?

maxmynter (approved):
Looks good to me. 

How do we want to handle integration tests? Should we make a separate issue for them? As the integration libraries are quite big, shall we use a separate `requirements-test.txt` or just add the dependecies to `requirements-dev.txt`?

src/lakefs_spec/spec.py
```diff
@@ -572,29 +572,35 @@ def ls(
...
+        Dispatch a lakeFS file-like object (local buffer on disk) for the given remote path for up- or downloads depending on ``mode``.
 
         Parameters
         ----------
         path: str | os.PathLike[str]
             The remote path for which to open a local ``LakeFSFile``. Must be a fully qualified lakeFS URI.
-        mode: Literal["rb", "wb", "xb"]
-            The file mode indicating its purpose. Use ``rb`` for downloads from lakeFS, ``wb/xb`` for uploads to lakeFS.
+        mode: Literal["r", "rb", "rt", "w", "wb", "wt", "x", "xb", "xt"]
+            The file mode indicating its purpose. Use ``r/rb`` for downloads from lakeFS, ``w/wb/x/xb`` for uploads to lakeFS.
```
maxmynter, line 593:
```suggestion
            The file mode indicating its purpose. Use ``r/rb`` for downloads from lakeFS, ``w/wb/x/xb`` for uploads to lakeFS. Note: Text modes are Note: Text modes are internally treated as binary modes.
```

nicholasjng: I'm not in favor of this anymore, as the text mode is coerced to the corresponding normal mode (`rt->r`, `wt->w`,...), which is not as surprising as binary coercion (in fact, that is exactly the same behavior as in builtin open()).

Should we just roll with that without a note?

PR #251, opened 2024-01-10
Add integration tests for third-party data science libraries

nicholasjng:
This is another smoke test to detect breakages in file system APIs. Adds pandas, polars, and duckdb to the test requirements.


nicholasjng:
This is failed config autodiscovery.

PR #260, opened 2024-01-27
Fix ls

renesat:
- Fix bag: return empty list if contain one directory
- Fix bag: return empty list when dir is not cached but have key in cache

I found it when working with DuckDB. To reproduce the second bug, you can try calling this:
```python
import duckdb
import fsspec

# Working
fsspec.filesystem("lakefs").glob("lakefs://test/main/test_dir/*.parquet")

# Not working
duckdb.register_filesystem(fsspec.filesystem("lakefs"))
duckdb.sql("FROM 'lakefs://test/main/test_dir/*.parquet'")

```

### DIFF
--- a/src/lakefs_spec/spec.py
+++ b/src/lakefs_spec/spec.py
@@ -379,7 +379,7 @@ class LakeFSFileSystem(AbstractFileSystem):
 
     def _update_dircache(self, info: list) -> None:
         """Update logic for dircache (optionally recursive) based on lakeFS API response"""
-        parents = {self._parent(i["name"]) for i in info}
+        parents = {self._parent(i["name"].rstrip("/")) for i in info}
         for pp in parents:
             # subset of info entries which are direct descendants of `parent`
             dir_info = [i for i in info if self._parent(i["name"].rstrip("/")) == pp]
@@ -485,7 +485,7 @@ class LakeFSFileSystem(AbstractFileSystem):
         list[str] | list[dict[str, Any]]
             A list of all objects' metadata under the given remote path if ``detail=True``, or alternatively only their names if ``detail=False``.
         """
-        path = stringify_path(path)
+        path = self._strip_protocol(path)
         repository, ref, prefix = parse(path)
 
         recursive = kwargs.pop("recursive", False)
@@ -540,7 +540,7 @@ class LakeFSFileSystem(AbstractFileSystem):
         # Retry the API call with appended slash if the current result
         # is just a single directory entry only (not its contents).
         # This is useful to allow `ls("repo/branch/dir")` calls without a trailing slash.
-        if len(info) == 1 and info[0]["type"] == "directory":
+        if len(info) == 1 and info[0]["type"] == "directory" and info[0]["name"] == path + "/":
             return self.ls(
                 path + "/",
                 detail=detail,
--- a/tests/test_ls.py
+++ b/tests/test_ls.py
@@ -39,6 +39,46 @@ def test_ls_caching(fs: LakeFSFileSystem, repository: Repository) -> None:
     assert counter.count("objects_api.list_objects") == 1
 
 
+def test_ls_with_one_dir(
+    fs: LakeFSFileSystem,
+    repository: Repository,
+    temp_branch: Branch,
+) -> None:
+    """
+    Check that ls calls are properly cached.
+    """
+    root = f"lakefs://{repository.id}/{temp_branch.id}"
+
+    fs.mkdir(f"{root}/test")
+    fs.cp(f"{root}/lakes.parquet", f"{root}/test/lakes.parquet")
+    fs.cp(f"{root}/lakes.parquet", f"{root}/test/lakes2.parquet")
+    fs.rm(f"{root}/README.md")
+    fs.rm(f"{root}/data/", recursive=True)
+    fs.rm(f"{root}/images/", recursive=True)
+    fs.rm(f"{root}/lakes.parquet")
+
+    # Check root
+    root_resource = f"{root}/"
+    list_of_files = fs.ls(root_resource)
+    assert [fs.unstrip_protocol(x["name"]) for x in list_of_files] == [f"{root_resource}test/"]
+    assert len(fs.dircache) == 1
+    assert {fs.unstrip_protocol(x) for x in fs.dircache.keys()} == {root_resource.removesuffix("/")}
+
+    # Check testdir
+    testdir = "test"
+    resource = f"{root}/{testdir}"
+    list_of_files = fs.ls(resource)
+    assert [fs.unstrip_protocol(x["name"]) for x in list_of_files] == [
+        f"{resource}/lakes.parquet",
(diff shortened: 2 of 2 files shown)


src/lakefs_spec/spec.py
```diff
@@ -542,7 +542,7 @@ def ls(
         # Retry the API call with appended slash if the current result
         # is just a single directory entry only (not its contents).
         # This is useful to allow `ls("repo/branch/dir")` calls without a trailing slash.
-        if len(info) == 1 and info[0]["type"] == "directory":
+        if len(info) == 1 and info[0]["type"] == "directory" and info[0]["name"] == path + "/":
```
nicholasjng, line 545: Unless I'm missing something, this should always be true, since `path` was requested in the original `reference.objects()` call (L520 above).

renesat: Not always. Only when we request path without "/" at the end. If we have only one directory in the requested path, it would be false.

renesat:
@nicholasjng  this is more complex example:
```python
import duckdb
import fsspec
import lakefs

lakefs.Repository("quickstart").create("local://quickstart", include_samples=True, exist_ok=True)

fs = fsspec.filesystem("lakefs")
fs.mkdir("lakefs://quickstart/main/test")
fs.cp("lakefs://quickstart/main/lakes.parquet", "lakefs://quickstart/main/test/lakes.parquet")
fs.cp("lakefs://quickstart/main/lakes.parquet", "lakefs://quickstart/main/test/lakes2.parquet")
fs.rm("lakefs://quickstart/main/README.md")
fs.rm("lakefs://quickstart/main/data/",  recursive=True)
fs.rm("lakefs://quickstart/main/images/",  recursive=True)
fs.rm("lakefs://quickstart/main/lakes.parquet")

# Bug 1: return empty list but have "test" dir
try:
    test1 = fsspec.filesystem("lakefs").ls("lakefs://quickstart/main/")
    print(test1)
    assert test1 != []
except:
    print("test1 not working")


# Bug 2: return empty list but have "test" dir

# Not working
try:
    test2 = fsspec.filesystem("lakefs").glob("lakefs://quickstart/main/test/*.parquet")
    print(test2)
    assert test2 != []
except:
    print("test2 not working")


duckdb.register_filesystem(fsspec.filesystem("lakefs"))

# Not working
try:
    test3 = duckdb.sql("FROM read_parquet('lakefs://quickstart/main/test/*.parquet')")
    print(test3)
except:
    print("test3 not working")

# Working
test4 = duckdb.sql("FROM read_parquet(['lakefs://quickstart/main/test/lakes2.parquet', 'lakefs://quickstart/main/test/lakes2.parquet'])")
print(test4)
```

I think bag 2 related with bag 1.

nicholasjng:
Sorry for the delay, and thanks for that new repro.
I ran this locally and was able to reproduce test case 1. We should therefore add it as a regression test. Would you like to do that?

Otherwise, looks good!

nicholasjng:
The test failure happens because the `path` is not stripped of its leading protocol in `test_ls_on_commit`.

Changing L488 in `fs.ls()` to `path = self._strip_protocol(path)` solves the issue.

renesat:
@nicholasjng I add test and find out why I have problem with line 430. It's because of incorrect cache update for directory with only directories in it.

src/lakefs_spec/spec.py
```diff
@@ -486,6 +486,7 @@ def ls(
             A list of all objects' metadata under the given remote path if ``detail=True``, or alternatively only their names if ``detail=False``.
         """
         path = stringify_path(path)
+        path = self._strip_protocol(path)
```
nicholasjng, line 489: `stringify_path()` is implied by `_strip_protocol()`, so you can replace L488 entirely.

tests/test_ls.py
```diff
@@ -38,6 +38,49 @@ def test_ls_caching(fs: LakeFSFileSystem, repository: Repository) -> None:
...
+    fs.rm(f"{root}/data/", recursive=True)
+    fs.rm(f"{root}/images/", recursive=True)
+    fs.rm(f"{root}/lakes.parquet")
+
+    # Check root
+    root_resource = f"{root}/"
+    list_of_files = fs.ls(root_resource)
+    assert [fs.unstrip_protocol(x["name"]) for x in list_of_files] == [
+        f"{root_resource}test/"
+    ]
```
nicholasjng, line 64: Seems like this fits on one line, could you run `pre-commit` to fix the formatting?

nicholasjng (commented):
Thanks, almost there!

tests/test_ls.py
```diff
@@ -39,6 +39,48 @@ def test_ls_caching(fs: LakeFSFileSystem, repository: Repository) -> None:
...
+    fs.rm(f"{root}/lakes.parquet")
+
+    # Check root
+    root_resource = f"{root}/"
+    list_of_files = fs.ls(root_resource)
+    assert [fs.unstrip_protocol(x["name"]) for x in list_of_files] == [f"{root_resource}test/"]
+    assert len(fs.dircache) == 1
+    assert set(map(lambda x: fs.unstrip_protocol(x), fs.dircache.keys())) == {
+        root_resource.removesuffix("/")
+    }
```
AdrianoKF, line 67: For consistent style (the rest of the file uses comprehensions instead of `map`):
```suggestion
    assert {fs.unstrip_protocol(x) for x in fs.dircache.keys()} == {root_resource.removesuffix("/")}
```

tests/test_ls.py
```diff
@@ -39,6 +39,48 @@ def test_ls_caching(fs: LakeFSFileSystem, repository: Repository) -> None:
...
+    # Check testdir
+    testdir = "test"
+    resource = f"{root}/{testdir}"
+    list_of_files = fs.ls(resource)
+    assert [fs.unstrip_protocol(x["name"]) for x in list_of_files] == [
+        f"{resource}/lakes.parquet",
+        f"{resource}/lakes2.parquet",
+    ]
+    assert len(fs.dircache) == 2
+    assert set(map(lambda x: fs.unstrip_protocol(x), fs.dircache.keys())) == {
```
AdrianoKF, line 78: See above:
```suggestion
    assert {fs.unstrip_protocol(x) for x in fs.dircache.keys()} == {
```

src/lakefs_spec/spec.py
```diff
@@ -427,7 +427,7 @@ def _ls_from_cache(self, path: str, recursive: bool = False) -> list[dict[str, A
                 result = []
             result.extend(files)
         if not result:
-            return result
+            return None
```
AdrianoKF, line 430: This change incurs an additional API call when the dircache returns an empty directory (`result == []`).

I would revert back to the old version, or make the intent more explicit by changing the condition to `if result is None`.

renesat: Yes, I forgot. It's the first attempt to fix the cache problem.

nicholasjng (approved):

nicholasjng:
Thank you!

PR #261, opened 2024-01-27
Split lakefs-spec test job into multiple jobs

nicholasjng:
This way, it is easier to understand whether a CI failure is a lint, test, or docs build failure. Previously, this was not possible as the pipeline did not tolerate failure.

--------------------------------

[Bump developer dependencies using pip compile --upgrade](https://github.com/aai-institute/lakefs-spec/commit/67cf9b6d0dd58e019bb3e4601f0be0da39446acd)

This addresses a number of security advisories related to Jupyter.

Since we do not distribute Jupyter code or expose any instances, these are not as important, but we can still bump the dependencies to silence the dependabot warnings.


nicholasjng:
Unfortunately, service containers do not work on non-Ubuntu Github runners, so we cannot currently test lakefs-spec on Windows and MacOS in CI.

All the more reason to get a deployment recipe going!

nicholasjng:
This cuts down CI runtime from about 2m30s minutes to around 1m30s. The CodeQL job with above two minutes average runtime is the bottleneck now, but this is still a nice improvement in my opinion.

cc @janwillemkl

nicholasjng:
I'm going to check this in unsquashed to a) modularize the CI pipeline asap and b) shorten the diff in the transaction redesign PR.

PR #266, opened 2024-02-21
Log a warning for uncommitted changes on transaction exit

### LINKED ISSUE #264: Warning message when ending a transaction while having uncommited objects
### What is the motivation and/or use case?

Consider the following scenario: A user starts a new transaction and adds an object to lakeFS. Without making a commit (forgetting to add it), the user ends the transaction.

A temporary branch is created for the scope of the transaction. The objects are uploaded to the temporary branch. Without a commit, there is nothing to merge back into the source branch, and the temporary branch is deleted.

This means that there is no 'evidence' left in lakeFS that any objects were uploaded. This may lead to accidental data loss, and without any messages or indication within lakeFS, it is very difficult to debug.

Showing the user a warning that objects were uploaded, but not committed, could help a great deal.

### How can we implement this feature?

_No response_

maxmynter:
Closes #264

### DIFF
--- a/src/lakefs_spec/transaction.py
+++ b/src/lakefs_spec/transaction.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 import logging
 import random
 import string
+import warnings
 from collections import deque
 from typing import TYPE_CHECKING, Literal, TypeVar
 
@@ -131,6 +132,12 @@ class LakeFSTransaction(Transaction):
 
         self.fs._intrans = False
         self.fs._transaction = None
+
+        if any(self._ephemeral_branch.uncommitted()):
+            msg = f"Finished transaction on branch {self._ephemeral_branch.id!r} with uncommitted changes."
+            if self.delete != "never":
+                msg += " Objects added but not committed are lost."
+            warnings.warn(msg)
 
         if success and self.automerge:
             if any(self.base_branch.diff(self._ephemeral_branch)):
--- a/tests/test_transactions.py
+++ b/tests/test_transactions.py
@@ -146,3 +146,33 @@ def test_transaction_bad_repo(fs: LakeFSFileSystem) -> None:
     with pytest.raises(ValueError, match="repository .* does not exist"):
         with fs.transaction(repository="REEEE"):
             pass
+
+
+def test_warn_uncommitted_changes(
+    random_file_factory: RandomFileFactory,
+    fs: LakeFSFileSystem,
+    repository: Repository,
+    temp_branch: Branch,
+) -> None:
+    random_file = random_file_factory.make()
+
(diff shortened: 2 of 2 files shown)


src/lakefs_spec/transaction.py
```diff
@@ -132,6 +132,11 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         self.fs._intrans = False
         self.fs._transaction = None
 
+        if any(True for _ in self._ephemeral_branch.uncommitted()):
+            logger.warning(
+                f"Transaction closed with uncommitted changes. {'Objects added but not committed are lost.' if self.delete != 'never' else ''}"
+            )
```
nicholasjng, line 138:
```suggestion
        if any(self._ephemeral_branch.uncommitted()):
            msg = f"Finished transaction {self._ephemeral_branch.id!r} contains uncommitted changes."
            if self.delete != "never":
                msg += " Objects added but not committed are lost."
            warnings.warn(msg)
```

Explanation: We always prefer a warning over a log, because that works even when the user does not set up logging. If the user needs logging, he can do `import logging; logging.capturewarnings(True)`, if he wants to catch the warnings, he can do so with a warnings filter context manager or by redirecting stderr.

Bonus: We keep a clean separation between stdout and stderr.

tests/test_transactions.py
```diff
@@ -146,3 +147,22 @@ def test_transaction_bad_repo(fs: LakeFSFileSystem) -> None:
...
+    random_file = random_file_factory.make()
+
+    lpath = str(random_file)
+
+    with caplog.at_level(logging.WARNING):
+        with fs.transaction(repository, temp_branch) as tx:
+            fs.put_file(lpath, f"{repository.id}/{tx.branch.id}/{random_file.name}")
+        assert len(caplog.records) == 1
+        assert caplog.records[0].levelname == "WARNING"
+        assert "uncommitted changes" in caplog.text
```
nicholasjng, line 168: Switch this to a `with pytest.warns(...)` with an appropriate regex. You might have to explicitly test `delete = "never"` to make the codecov bot happy.

src/lakefs_spec/transaction.py
```diff
@@ -132,6 +133,12 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         self.fs._intrans = False
         self.fs._transaction = None
 
+        if any(True for _ in self._ephemeral_branch.uncommitted()):
```
nicholasjng, line 136:
```suggestion
        if any(self._ephemeral_branch.uncommitted()):
```

nicholasjng (commented):
Nice work!

nicholasjng (approved):

PR #268, opened 2024-02-28
Docs: Fix code highlighting in third-party libraries integration guide

### LINKED ISSUE #267: Docs: Code highlighting in Polars integration example is offset
### What happened?

The line highlighting for the code in the [Polars integration example](https://lakefs-spec.org/latest/guides/integrations/#polars) is shifted and the numbered annotation is dangling apparently:

image

### What did you expect to happen?

_No response_

### lakeFS-spec version

0.7.0

### lakeFS version

n/a

AdrianoKF:
After a refactoring of the transactions feature, the code examples themselves were modied, but not the page that highlights some lines in them.

This PR updates those code blocks to make them consistent with the actual example code.

Additionally, it pins the Python version used for pre-commit hooks to 3.11, to prevent problems on systems with Python 3.12 (which would be used to lock dependencies using `pip-compile`, which includes the Python version number in a comment in the generated dependency lock file).

Closes #267

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,5 @@
+default_language_version:
+  python: python3.11
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.5.0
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -13,27 +15,27 @@ repos:
...
       - id: ruff-format
   - repo: https://github.com/PyCQA/bandit
     rev: 1.7.7
     hooks:
-    - id: bandit
-      args: [-c, pyproject.toml]
-      additional_dependencies: ["bandit[toml]"]
+      - id: bandit
+        args: [-c, pyproject.toml]
+        additional_dependencies: ["bandit[toml]"]
```
maxmynter, line 32: I agree with these indents here. But Nicho noted them on another PR from me. Do we have a projet standard on these idents/formatting things?

maxmynter (approved):

.pre-commit-config.yaml
janwillemkl: The rest of the file uses this style of indentation. From a quick look we also use it in the other files.

AFAIK, we do not have a standard defined.

AdrianoKF: We don't, but at least each file should be consistent. FWIW, I consider whatever defaults the YAML plugin in VS Code has my standard (which is why the file got reformatted when I saved it).

PR #271, opened 2024-03-07
Fixed broken URL in README.md

leonpawelzik:
Replaced the old link(s) within the README.md that caused a 404, with the correct user guide and tutorial link.


AdrianoKF (approved):
Good catch!

PR #272, opened 2024-03-15
Add links to subpages in lakefs_spec API reference

leonpawelzik:
## Add subpage links to lakefs_spec API reference

This pull request enhances the `lakefs_spec` API reference page by adding links to the following subpages:

- `errors`
- `spec`
- `transaction`
- `util`

The `gen_api_ref_pages.py` script has been updated to conditionally include a text block with these links when generating the `lakefs_spec` page. This improves navigation and discoverability of the subpages within the API reference.

Resolves: Issue https://github.com/aai-institute/MLOps-Engineering/issues/55 .


nicholasjng:
Please `git rm docs/.DS_Store` again, and consider putting it into your machine's global gitignore: https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

nicholasjng (approved):
The `spec` point description really makes me think that we should have renamed the `spec` module to `lakefs_spec.core` instead, like s3fs and gcsfs do.

With today's git blame following renames, maybe that's still a sensible thing to do.

PR #273, opened 2024-03-20
Docs: Link to root-level modules in API docs

AdrianoKF:
This adds functionality to `gen_api_ref_pages.py` to extract the module docstrings for root-level modules in the package, and adds links to them to the index page of the API reference.

image

See also previous PR: #272


leonpawelzik (approved):

nicholasjng:
Once the visual confirmation is there for nnbench, we can go ahead with this.

If you feel super instructive, you can give a quick comment about how `__init__.py`s contain submodule docstrings (i.e. collections of .py files under its tree) as opposed to normal .py files, but I'm happy either way.

PR #274, opened 2024-03-29
fs.exists() should return True for common prefixes

ozkatz:
Similar to s3fs, `fs.exists()` should return True also for common prefixes.
Ran into this while trying to use lakefs-spec with [HuggingFace Datasets](https://huggingface.co/docs/datasets/en/index) - loading a dataset accepts a directory uri, which is checked for existence and subsequently fails.

The check is done in a similar fashion to [s3fs'](https://github.com/fsspec/s3fs/blob/main/s3fs/core.py#L1397) - ensure path ends with `'/'`, attempt listing, return True if there's anything under that prefix.

### DIFF
--- a/src/lakefs_spec/spec.py
+++ b/src/lakefs_spec/spec.py
@@ -236,7 +236,13 @@ class LakeFSFileSystem(AbstractFileSystem):
         repository, ref, resource = parse(path)
         try:
             reference = lakefs.Reference(repository, ref, client=self.client)
-            return reference.object(resource).exists()
+            if reference.object(resource).exists():
+                return True
+            # if it isn't an object, it might be a common prefix (i.e. "directory").
+            children = reference.objects(
+                max_amount=1, prefix=resource.rstrip("/") + "/", delimiter="/"
+            )
+            return len(list(children)) > 0
         except ServerException as e:
             # in case of an error other than "not found", existence cannot be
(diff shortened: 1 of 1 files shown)


ozkatz:
@nicholasjng @AdrianoKF Hope this change meets the contribution guidelines, happy to change as needed.

nicholasjng:
Looks good! I'm only wondering about exceptions - can the exists() call throw in any non-trivial (i.e. other than connection/permission errors) situation, meaning we wouldn't get to decide the prefix question?

ozkatz:
Thanks @nicholasjng - good question :) I don't think calling `objects()` increases the surface area for that. 

You're already handling server errors, which is the right thing to do - this should also handle any unexpected errors from listing objects.

nicholasjng (approved):
Nice! Please squash to a single commit, then we can get this in :)

ozkatz:
@nicholasjng done :)

nicholasjng:
Thank you!

ozkatz:
@nicholasjng any estimation as to when we can expect a next release?

nicholasjng:
@ozkatz I can release today if it's urgent, but if anything goes wrong, it takes until Tuesday at minimum.

ozkatz:
Tuesday is absolutely fine :) thank you @nicholasjng !!

PR #275, opened 2024-04-03
Docs: Demonstrate integration of Hugging Face Datasets

AdrianoKF:
Similar to the [lakeFS PR](https://github.com/treeverse/lakeFS/pull/7613), this PR adds an example of how to interact with lakeFS repositories from the HF Datasets library.

It also adds a mention of the library in the repository README and docs landing page.

### DIFF
--- a/docs/guides/integrations.md
+++ b/docs/guides/integrations.md
@@ -38,6 +38,18 @@ Similar to the example above, the following code snippet illustrates how to read
 
 1. Makes the lakeFS-spec file system known to DuckDB (`duckdb.register_filesystem(fsspec.filesystem("lakefs"))` can also be used to avoid the direct import of `LakeFSFileSystem`)
 
+## Hugging Face Datasets
+
+[Hugging Face :hugging_face: Datasets](https://huggingface.co/docs/datasets/index){: target="_blank" rel="noopener"} is a library for easily accessing and sharing datasets for Audio, Computer Vision, and Natural Language Processing (NLP) tasks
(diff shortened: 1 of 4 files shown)


nicholasjng (approved):
Nice! 

On a meta-level: I remember that the polars code example broke out of the blue in some newer version. Should we add those examples to an integration test suite?

docs/guides/integrations.md
```diff
@@ -38,6 +38,18 @@ Similar to the example above, the following code snippet illustrates how to read
 
 1. Makes the lakeFS-spec file system known to DuckDB (`duckdb.register_filesystem(fsspec.filesystem("lakefs"))` can also be used to avoid the direct import of `LakeFSFileSystem`)
 
+## Hugging Face Datasets
+
+[Hugging Face :hugging_face: Datasets](https://huggingface.co/docs/datasets/index){: target="_blank" rel="noopener"} is a library for easily accessing and sharing datasets for Audio, Computer Vision, and Natural Language Processing (NLP) tasks
```
nicholasjng, line 43:
```suggestion
[Hugging Face :hugging_face: Datasets](https://huggingface.co/docs/datasets/index){: target="_blank" rel="noopener"} is a library for easily accessing and sharing datasets for Audio, Computer Vision, and Natural Language Processing (NLP) tasks.
```

AdrianoKF:
> Nice!
> 
> On a meta-level: I remember that the polars code example broke out of the blue in some newer version. Should we add those examples to an integration test suite?

Good point. Eventually, I'd very much like that - the only problem is that the code examples in the docs currently assume a certain environment (but I guess we can hardcode that assumption for the time being, still better than no tests at all).

PR #278, opened 2024-04-29
Update pre-commit hooks, change lockfile to use uv pip compile

nicholasjng:
uv is much faster and thus more appropriate for a mandatory commit hook.

NB: setuptools is now in the lockfile, because uv does not have a "unsafe package exclusion" facility.


AdrianoKF:
Can you also update `hack/lock-deps.sh`, please, @nicholasjng?

nicholasjng:
So codecov fails the pipeline every time code is deleted from the repo because of the percentage changes? ¯\\\_(ツ)\_/¯

AdrianoKF:
> So codecov fails the pipeline every time code is deleted from the repo because of the percentage changes? ¯\_(ツ)_/¯

It's set up to prevent PRs with decreasing coverage atm. It seems that coverage actually went down in `errors.py`: https://app.codecov.io/gh/aai-institute/lakefs-spec/pull/278/blob/src/lakefs_spec/errors.py

image

It seems that one of the dependency updates might have made that `else` branch obsolete.

AdrianoKF:
I've configured Codecov to allow PRs that decrease coverage by 1% at most in 0f7b811050fb4caf1b85954984bfb8667ea1f94d.

As discussed on Slack, the check in `errors.py` is still valid for older versions of the lakeFS SDK that don't include the fix in https://github.com/treeverse/lakeFS/pull/7697.

PR #280, opened 2024-05-06
Bump minimum fsspec version to 2023.12.0

nicholasjng:
This means that the polymorphic filesystem facility, which was upstreamed to fsspec and first appeared in that release, can go away.

----------------

Contains the error handling fix from #278 as a cherrypick.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,7 +34,7 @@ classifiers = [
     "Typing :: Typed",
 ]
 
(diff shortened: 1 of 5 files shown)


AdrianoKF (approved):

AdrianoKF:
We'll have to see about the changes to coverage in the error handling code, but I'd leave that up to the other PR #278.

Issue #279, opened 2024-04-30
Async support?


ion-elgreco:
### What is the motivation and/or use case?

Is async supported in lakefs spec?

AdrianoKF:
Thanks for raising this issue, @ion-elgreco!

Currently, there is no async support in lakeFS-spec, since the lakeFS client does not support it either (only for import operations, which are outside of the scope of lakeFS-spec).
We might consider this feature in the future depending on the overall demand.

Could you elaborate a bit on your use case?

ion-elgreco:
@AdrianoKF I need to read a couple of hundred thousands text files and parse them with regex. Doing this asynchronous is much faster.

I have currently written it with s3fs and lakefs S3 gateway. Seeing 15-20x performance increase versus sequential reads

leonpawelzik:
Hey @ion-elgreco,

would you be willing to test out our WIP solution: https://github.com/aai-institute/lakefs-spec/pull/283?

ion-elgreco:
@leonpawelzik currently on holidays, and afterwards I won't be using LakeFS anytime soon.

Might be an idea to create small benchmark that compares s3fs vs lakefs-spec async perf

PR #283, opened 2024-07-05
Async filesystem implementation

AdrianoKF:
The implementation simply wraps the synchronous file system already present in the code in `asyncio` coroutines.

Might serve as the first implementation for #279.

### DIFF
--- a/src/lakefs_spec/util.py
+++ b/src/lakefs_spec/util.py
@@ -4,13 +4,16 @@ Useful utilities for handling lakeFS URIs and results of lakeFS API calls.
 
 from __future__ import annotations
 
+import asyncio
+import functools
 import hashlib
 import os
 import re
-from typing import Any, Callable, Generator, Protocol
+from typing import Any, Callable, Coroutine, Generator, Protocol, TypeVar
 
 from lakefs_sdk import Pagination
 from lakefs_sdk import __version__ as __lakefs_sdk_version__
+from typing_extensions import ParamSpec
 
 lakefs_sdk_version = tuple(int(v) for v in __lakefs_sdk_version__.split("."))
 del __lakefs_sdk_version__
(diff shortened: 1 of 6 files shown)


src/lakefs_spec/util.py
```diff
@@ -4,13 +4,16 @@
...
+import functools
 import hashlib
 import os
 import re
-from typing import Any, Callable, Generator, Protocol
+from typing import Any, Callable, Coroutine, Generator, Protocol, TypeVar
 
 from lakefs_sdk import Pagination
 from lakefs_sdk import __version__ as __lakefs_sdk_version__
+from typing_extensions import ParamSpec
```
nicholasjng, line 16: This is Python>=3.10 territory, so we should mark our package as that. (It is about time, anyway.)

nicholasjng (commented):
Is there an advantage to implementing async in this way? I seem to remember s3fs doing the opposite, i.e. implementing all methods as async and then applying a `sync_wrapper` on top of it for the syncs.

They inherit only from `AsyncFileSystem`: https://github.com/fsspec/s3fs/blob/main/s3fs/core.py#L176

Happy for a discussion.

src/lakefs_spec/util.py
AdrianoKF: That's why it's imported from `typing_extensions`, so we can keep 3.9-compatibility.

AdrianoKF:
> Is there an advantage to implementing async in this way? I seem to remember s3fs doing the opposite, i.e. implementing all methods as async and then applying a `sync_wrapper` on top of it for the syncs.
> 
> They inherit only from `AsyncFileSystem`: https://github.com/fsspec/s3fs/blob/main/s3fs/core.py#L176
> 
> Happy for a discussion.

Generally, implementing only the async version and then wrapping it to make it blocking would be the more correct way, IMO. The problem is that the lakeFS high-level SDK does not expose an async API (and we rewrote our implementation based on that). If we wanted to go async-first, we have two options:

1. Suggest a feature to make the high-level SDK non-blocking (which would be great, but unlikely to happen, IMO, because it would subvert the strong typing guarantees of the SDK, in the end leading to two parallel APIs).
2. Revert our code back to the low-level SDK (i.e., the OpenAPI-generated API client), which supports [async API requests with `async_req=True`](https://github.com/treeverse/lakeFS/blob/master/clients/python/lakefs_sdk/api_client.py#L358-L364).
3. (okay, three options) keep the sync->async wrapper for and use it as a benchmarking baseline, so we can evaluate the other two options going forward.

PR #287, opened 2024-10-01
Update project lockfiles and pre-commit setup

nicholasjng:
Running `pre-commit autoupdate` and `uv pip compile ... --upgrade`.

Due to behavior quirks in newest pydoclint (v0.5.9), disable the hook for now, until a way to deal with this has been found (I raised an issue on the repo today).


nicholasjng:
@AdrianoKF Since we're not adding any more APIs in the foreseeable future, I opted to disable pydoclint for now until the DOC503 rule stabilizes. I could alternatively pin it to an older version.

Issue #291, opened 2024-10-09
fs.exists() on the root of a branch throws an exception


ssbusc1:
### What happened?

I have a LakeFS repository 'demo1', and a 'main' branch. I expect `lakefs://demo1/main` to represent the root of a filesystem corresponding to that branch. However an `exists("lakefs://demo1/main")`, with or without a trailing "/" throws an exception.

Example test:
```
from lakefs_spec import LakeFSFileSystem

def test_lakefs_exists():
    fs = LakeFSFileSystem( )
    assert fs.exists("lakefs://demo1/main/")  # throws OSError
```
../.venv/lib/python3.12/site-packages/lakefs_spec/spec.py:233: OSError

System details:
Python 3.12
pip freeze | grep -i 'lakefs\|fsspec'
fsspec==2024.6.1
lakefs==0.7.1
lakefs-sdk==1.37.0
lakefs-spec==0.10.0

### What did you expect to happen?

If the repo and ref exist, then the root of the filesystem at that ref should always exist. As a reference, the [S3FS implementation](https://github.com/fsspec/s3fs/blob/main/s3fs/core.py#L1056-L1057) always returns true for "" and "/" provided as the path being checked.

The current lakefs-spec implementation seems to defer the check to the backend as an object check. I'm not sure if that is actually required, instead of checking the repo + ref exist and always returning `true` as in the S3FS implementation.

Note that this behavior is also inconsistent with the behavior of `fs.ls()`, `fs.walk()`, etc.:
```
from lakefs_spec import LakeFSFileSystem

def test_lakefs_ls():
    fs = LakeFSFileSystem()

    path = "lakefs://demo1/main/"
    files = fs.ls(path)  # No exception
    assert len(files) > 0
```

### lakeFS-spec version

Version: 0.10.0

### lakeFS version

Version: 0.7.1

AdrianoKF:
Hi @ssbusc1,

Thanks for reporting the issue and providing a thorough description!

We have implemented a fix in #292, I will shortly release a new version of the package on PyPI.

`fs.exists("lakefs://repo/<ref>/")` will be supported and basically check if the specific ref (branch/commit) exists in the repository. Note that the trailing slash is required.

Feel free to ping me if you notice any other problems with the latest code!

ssbusc1:
Thanks @AdrianoKF for the quick fix and support! I'm now able to move ahead with version 0.11.0 and will report back if I run into any other issues.

PR #292, opened 2024-10-09
Support `fs.exists()` on repo root

AdrianoKF:
This PR implements `fs.exists()` on the root of a repository.

Semantically, checking the existence of a repo root equals checking if the specified ref (commit ID/branch name) exists in the repository.

Additionally, it fixes an inconsistency between the `batched` utility function between Python <3.12 and later versions and expands test coverage in CI to all supported Python versions.

Fixes issue #291


AdrianoKF:
Docs jobs in CI fails due to a renamed file, will be resolved by merging into `main`.

PR #294, opened 2024-10-09
Tests: Speed up `test_rm_with_1k_objects_or_more`

AdrianoKF:
By employing simple `asyncio` thread parallelism, the execution of the `test_rm_with_1k_objects_or_more` test is sped up from ~60s to around ~5s.

Unfortunately, this requires the `pytest-asyncio` plugin, but IMO the overall speedup is worth the extra dev dependency.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,7 +34,10 @@ classifiers = [
     "Typing :: Typed",
 ]
 
-dependencies = ["fsspec>=2023.12.0", "lakefs>=0.2.0"]
+dependencies = [
+    "fsspec>=2023.12.0",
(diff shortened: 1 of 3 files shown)


nicholasjng (approved):
Great stuff!

Why is `pytest-asyncio` required now? Is that just obligatory once you have an async test?

AdrianoKF (commented):
> Why is `pytest-asyncio` required now? Is that just obligatory once you have an async test? 

Yup, otherwise there's no asyncio event loop (I guess you could spawn your own as part of the test functions, but it might become iffy quickly).

PR #295, opened 2024-10-26
Add Python 3.13 to test matrix

nicholasjng:
Now that it's released, we should test it.

--------------

This will not pass until pyarrow (and possibly others) release Python 3.13 wheels. Until then, we could look at marking integration tests as xfail for 3.13, but not sure if that's worth the effort.

### DIFF
--- a/.github/workflows/python.yaml
+++ b/.github/workflows/python.yaml
@@ -30,7 +30,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python-version: ["3.10", "3.11", "3.12"]
+        python-version: ["3.10", "3.11", "3.12", "3.13"]
         deps:
           # Oldest supported versions
           - lakefs: "lakefs==0.2.0"
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,14 +10,14 @@ repos:
       - id: end-of-file-fixer
       - id: mixed-line-ending
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.11.2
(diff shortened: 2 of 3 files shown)


AdrianoKF:
Re [ef57f6f](https://github.com/aai-institute/lakefs-spec/pull/295/commits/ef57f6fa2b86598083ce02cf9cf94e3bba93d3d0): I'm not sure how I feel about bumping everything at once - does `uv lock -P pyarrow` suffice to make the tests pass already?

AdrianoKF:
Also: Should we add a version guard `requires-python: ">=3.9,<3.14"` to `pyproject.toml`, since this mirrors the versions we explicitly test against?

nicholasjng:
I decided on a full upgrade in place of trying to upgrade pandas, numpy, polars etc., but I can do that too.

AdrianoKF:
> I decided on a full upgrade in place of trying to upgrade pandas, numpy, polars etc., but I can do that too.

I guess if it doesn't break anything obvious it doesn't hurt (since we also don't constrain the versions in pyproject.toml).

AdrianoKF:
> > Also: Should we add a version guard requires-python: ">=3.9,<3.14" to pyproject.toml, since this mirrors the versions we explicitly test against?
> 
> Semantically, all current non-development Python versions fulfill this constraint, which makes it an empty condition (until a very brief transition time in October 2025). I've seen these constraints in practice, but I'm not convinced they are all that useful.

I actually meant `>=3.10`, but your point still stands somewhat. I guess what it does it give a "
proper" notice once Pie-thon 3.14 comes out since you won't be able to install the package without us at least giving it a cursory test beforehand.

Issue #297, opened 2024-12-13
Strange behavior with files containing `__` 


indigoviolet:
### What happened?

```
❯ lakectl fs ls -r lakefs://sample/test-venky/data/foo/
object          2024-12-12 23:32:03 -0800 PST    2.7 kB          data/foo/bar/axe.just
object          2024-12-12 23:26:00 -0800 PST    0 B             data/foo/bar/quux/widg.txt
object          2024-12-12 23:26:00 -0800 PST    0 B             data/foo/bar__baz.txt
```

```
p = UPath("lakefs://sample/test-venky/data/foo", use_listings_cache=False)
list(p.fs.find(p, withdirs=True))
---
['sample/test-venky/data/foo',
 'sample/test-venky/data/foo/bar/',
 'sample/test-venky/data/foo/bar__baz.txt']
```
 
Where did `foo/bar/*/` go? 

`p.fs.walk()` has the same behavior, but `p.fs.ls(p, recursive=True)` returns the right information

The files do exist if you drill down and look explicitly for them

```
p = UPath("lakefs://sample/test-venky/data/foo/bar/axe", use_listings_cache=False)
p.fs.ls(p)
---
[{'checksum': 'a87e010a318457fd39f9525e6d1c576d',
  'content-type': 'application/octet-stream',
  'mtime': 1734075123,
  'name': 'sample/test-venky/data/foo/bar/axe.just',
  'size': 2696,
  'type': 'object'}]
```

A further problem is that this leads to fs.walk reporting non-existent files:

```
p = UPath("lakefs://sample/test-venky/data/foo/", use_listings_cache=False)
list(p.fs.walk(p))
---
[('sample/test-venky/data/foo', ['bar'], ['bar__baz.txt']),
 ('sample/test-venky/data/foo/bar', [], ['', 'bar__baz.txt'])]
```

^ notice that bar__baz.txt is returned twice, once under foo/  and once under bar/ -- the second instance is just wrong

### What did you expect to happen?

The pattern seems to be that if you have if you have a folder x/ and a file x__<whatever> as siblings, the children of x/ are hidden somehow

### lakeFS-spec version

0.11.0

### lakeFS version

0.7.1

AdrianoKF:
Thanks for your detailed report, @indigoviolet! I'll investigate and get back to you soon.

AdrianoKF:
@indigoviolet, I've pushed a potential fix for the issue you are seeing in PR #298.

[...]

PR #298, opened 2024-12-16
Handle partial name matches in `ls` method

AdrianoKF:
This PR fixes the `ls` implementation to exclude partial (prefix-only) matches when listing objects in a repository.

The previous behavior would incorrectly include files with names that are a prefix of the path passed to `ls`, e.g., a file named `foo/bar_1` would be included when listing `foo/bar` (a directory).

The root cause is that the `GET /repositories/{repository}/refs/{ref}/objects/ls` lakeFS API call returns objects based on their prefix, not full name matches, so we have to perform additional filtering in our `ls` method.

A regression test has been added based on the issue description.

Issue: #297

### DIFF
--- a/src/lakefs_spec/spec.py
+++ b/src/lakefs_spec/spec.py
@@ -365,7 +365,7 @@ class LakeFSFileSystem(AbstractFileSystem):
             except ServerException as e:
                 raise translate_lakefs_error(e, rpath=path)
 
-        out = self.ls(path, detail=True, recursive=True, **kwargs)
+        out = self.ls(path, detail=True, recursive=False, **kwargs)
         if not out:
             raise FileNotFoundError(errno.ENOENT, os.strerror(errno.ENOENT), path)
 
@@ -504,8 +504,8 @@ class LakeFSFileSystem(AbstractFileSystem):
 
         kwargs["prefix"] = prefix
 
-        info = []
         # stat infos are either the path only (`detail=False`) or a dict full of metadata
+        info = []
         delimiter = "" if recursive else "/"
         reference = lakefs.Reference(repository, ref, client=self.client)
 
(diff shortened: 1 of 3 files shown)


src/lakefs_spec/spec.py
```diff
@@ -521,6 +521,15 @@ def ls(
                         }
                     )
                 elif isinstance(obj, ObjectInfo):
+                    # Skip over prefix-only matches, e.g.:
+                    # - foo/bar
+                    # -
```
nicholasjng, line 526: Does this empty bullet serve the explanation (in this case, I don't quite get it)?

nicholasjng (approved):

src/lakefs_spec/spec.py
```diff
@@ -521,6 +521,15 @@ def ls(
...
+                    # - foo/bar_test.txt
+
+                    # `ls("foo/bar")` should not include `bar_test.txt`
+                    if not (prefix == "" or prefix.endswith("/")) and obj.path != prefix:
```
nicholasjng, line 530: Note to self: This is equivalent to `if prefix and not prefix.endswith("/") and obj.path != prefix`.

Condition 1 is the repo root (which is always a directory), condition two is an explicit directory ls request, and condition 3 is a file stat request (like from `info()`).

So the condition can be thought of as "if the requested prefix does not look like a directory, and it does not equal the object path, we assume it's a file matching the prefix and ignore it", which is what we want.

Note that in the case of a directory without the terminating slash, we retry below (L547+) with the slash-terminated path.

src/lakefs_spec/spec.py
AdrianoKF: Thanks, that was a stray edit. I have improved the appearance in 83a9255.

nicholasjng (approved):

PR #300, opened 2024-12-18
Support a wider range of ref expressions

nicholasjng:
Adds dot ("."), caret ("^"), and tilde ("~") to the ref identifier regex portion, since they are allowed in ref expressions in lakeFS.

New test cases for each of these scenarios were added to the existing parsing test.

------------

~Also contains a round of `pre-commit autoupdate` that I checked in by mistake apparently.~

Ancestor statements were ~not yet~ verified to work on e.g. a lakeFS quickstart sidecar container, I plan to do that early tomorrow. (But allowing them probably won't hurt, simply because lakeFS mentions them as part of their ref spec.)

### DIFF
--- a/src/lakefs_spec/util.py
+++ b/src/lakefs_spec/util.py
@@ -121,7 +121,7 @@ def parse(path: str) -> tuple[str, str, str]:
     uri_parts = {
         "protocol": r"^(?:lakefs://)?",  # leading lakefs:// protocol (optional)
         "repository": r"(?P<repository>[a-z0-9][a-z0-9\-]{2,62})/",
-        "ref expression": r"(?P<ref>\w[\w\-]*)/",
+        "ref expression": r"(?P<ref>\w[\w\-.^~]*)/",
         "resource": r"(?P<resource>.*)",
     }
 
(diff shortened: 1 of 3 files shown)


AdrianoKF:
@nicholasjng, can you also add a regression test for the examples given in the issue?

I like the idea keeping those explicit, since they also provide traceability to the underlying issues. I've started doing the same in PR #298, see the `tests/regression` folder.

.pre-commit-config.yaml
```diff
@@ -17,13 +17,13 @@ repos:
...
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.7.1
+    rev: v0.8.3
     hooks:
       - id: ruff
         args: [--fix, --exit-non-zero-on-fix]
       - id: ruff-format
   - repo: https://github.com/PyCQA/bandit
-    rev: 1.7.10
+    rev: 1.8.0
```
AdrianoKF, line 26: Not urgent, but given that ruff implements the [`flake8-bandit` rules](https://docs.astral.sh/ruff/rules/#flake8-bandit-s), we might get rid of this hook.

nicholasjng: I reverted the whole update due to the observed lock file problems with setuptools_scm, but it's a good thing to note.

Issue #299, opened 2024-12-17
`util.parse` should support accessing tags, which would require a less restrictive regex


gaurav:
`util.parse()` currently assumes that the branch name is "Only letters, digits, underscores and dash, no leading dash."

https://github.com/aai-institute/lakefs-spec/blob/06134be76122834f197aba1e3fee89628e66236d/src/lakefs_spec/util.py#L118-L130

However, users might want to use [a tag name](https://docs.lakefs.io/understand/model.html#tags) instead of a branch name, which has a [much more complicated syntax](https://git-scm.com/docs/git-check-ref-format). Notably, the current regex doesn't match the two example tags in the LakeFS docs (`v2.3` and `dev-jane-before-v2.3-merge`). It might also be useful to support [ref expressions](https://docs.lakefs.io/understand/model.html#ref-expressions).

I would suggest making the regex less restrictive -- at a minimum `(\w[\w\-\.]*)`, but possibly going all the way to support ref expressions to include `[~^]`. The error could also be rewritten to be a bit clearer (e.g. because `lakefs.ls(...)` calls this after calling `_strip_protocol()`, I get the following error message: `ValueError: expected path with structure lakefs://<repo>/<ref>/<resource>, got 'heal-studies/v2.0/'` -- which made me think that I had forgotten to add the `lakefs://` prefix before realizing that the branch was being rejected for being incorrect).

Alternatively, if I'm misunderstanding how tags and ref expressions should be used in lakefs-spec, please let me know -- and thank you so much for building this great library!

nicholasjng:
Thank you for the detailed report! 

Since we want to fully implement lakeFS's URI specification, we must also match their support for ref expressions. Your regex suggestion sounds great, would you be willing to send a PR?

I agree that the error message is a bit sparse, and even misleading in the case you presented. We might need to check our callsites for the order of `util.parse()` and `_strip_protocol()` - alternatively, we could decide to split the protocol off the template path if the input doesn't have one, so we don't confuse people with the protocol unnecessarily.

gaurav:
This works great now, thanks so much for the quick fix!

PR #301, opened 2024-12-20
Improve error message in `util.parse()` by incrementally using regexes

nicholasjng:
This loses a bit of performance, but allows us to improve the error message to the user by attaching information about which part of the URI was invalid.

Since we consider the regexes an implementation detail, we do not return them in the error message.

Changes the failure test to include the expected invalid parts to harden the test case.

Part of #299.

### DIFF
--- a/src/lakefs_spec/util.py
+++ b/src/lakefs_spec/util.py
@@ -7,6 +7,7 @@ import itertools
 import os
 import re
 import sys
+from collections import namedtuple
 from collections.abc import Callable, Generator, Iterable, Iterator
 from typing import Any, Protocol
 
@@ -15,6 +16,8 @@ from lakefs_sdk import __version__ as __lakefs_sdk_version__
 
 lakefs_sdk_version = tuple(int(v) for v in __lakefs_sdk_version__.split("."))
 del __lakefs_sdk_version__
+
+_ParsedUri = namedtuple("_ParsedUri", ("repository", "ref", "resource"))
 
 
 class PaginatedApiResponse(Protocol):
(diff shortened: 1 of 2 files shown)


src/lakefs_spec/util.py
```diff
@@ -115,16 +118,25 @@ def parse(path: str) -> tuple[str, str, str]:
...
+    regexes = {
+        "protocol": r"^(?:lakefs://)?",  # leading lakefs:// protocol (optional)
+        "repository": r"(?P<repository>[a-z0-9][a-z0-9\-]{2,62})/",
+        "ref expression": r"(?P<ref>\w[\w\-]*)/",
+        "resource": r"(?P<resource>.*)",
+    }
+
+    groups: dict[str, str] = {}
+    start = 0
+    for group, regex in regexes.items():
```
AdrianoKF, line 130: Wouldn't mind a small comment here explaining the rationale for the iterative parsing.

src/lakefs_spec/util.py
```diff
@@ -115,16 +118,25 @@ def parse(path: str) -> tuple[str, str, str]:
...
-    # Second regex is the branch: Only letters, digits, underscores and dash, no leading dash.
-    path_regex = re.compile(r"(?:lakefs://)?([a-z0-9][a-z0-9\-]{2,62})/(\w[\w\-]*)/(.*)")
-    results = path_regex.fullmatch(path)
-    if results is None:
-        raise ValueError(
-            f"expected path with structure lakefs://<repo>/<ref>/<resource>, got {path!r}"
-        )
-
-    repo, ref, resource = results.groups()
+    regexes = {
```
AdrianoKF, line 121: Maybe `uri_parts`?

src/lakefs_spec/util.py
nicholasjng: Added a few lines just below the loop.

src/lakefs_spec/util.py
nicholasjng: Done.

AdrianoKF (approved):

PR #308, opened 2024-12-20
Use PEP735 dependency groups instead of extras

nicholasjng:
These are now supported by uv, and since they are dev-only (lakefs-spec does not contain any user-facing extra functionality), we can move them to PEP735 dependency groups instead.


AdrianoKF:
Sorry for post-merge review, but we'll also need to adapt the contribution guide, which mentions the installation of the extra dependencies, IIRC.

PR #310, opened 2024-12-27
Update `setup-uv` action to v5, tweak install and setup

nicholasjng:
v5 of `setup-uv` adds caching and a default venv, so things should be a bit faster on repeated runs.

Since `python-version` allows bootstrapping a Python version directly on action setup, we do that instead.

### DIFF
--- a/.github/actions/python-deps/action.yaml
+++ b/.github/actions/python-deps/action.yaml
@@ -8,10 +8,9 @@ runs:
   using: "composite"
   steps:
     - name: Install uv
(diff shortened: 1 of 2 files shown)


nicholasjng:
I'm still not sure when we need to sync and when not to - judging by the docs, `uv run pytest` seems to work even without a prior sync. Maybe the project is already installed on venv creation?

I'll merge this, but it seems we could squeeze out more perf in some places.

Issue #314, opened 2025-01-16
Feature Request: Enable `@` and `~N` Syntax in `lakefs-spec` Paths


grepinsight:
### What is the motivation and/or use case?

The [performance best practices documentation](https://docs.lakefs.io/understand/performance-best-practices.html#read-data-using-the-commit-id) highlights the advantages of accessing committed data using syntax like `lakefs://repo/main@/path`. Specifically:

> In cases where you are only interested in reading committed data:  
> Add `@` before the path `lakefs://repo/main@/path`.  
> When accessing data using the branch name (e.g., `lakefs://repo/main/path`), lakeFS will also try to fetch uncommitted data...

This feature works seamlessly with `lakectl`. However, `lakefs-spec` currently does not support the `@` or `~N` syntax in paths, resulting in the following error:

```python
ValueError: not a valid lakeFS URI: ...
```

### Desired Functionality

Extend `lakefs-spec` to support the `@` and `~N` syntax in paths, enabling users to directly specify:
1. **Exact commit reference:** Use `@` to specify a path as of a specific commit on a branch.
2. **Relative commit reference:** Use `~N` to access data from `N` commits prior to the latest commit on a branch.

For example:
```python
# Desired path syntax with `@` for a specific commit
file_path = "lakefs://example-repo/main@/data.parquet"

# Desired path syntax with `~N` for relative commits
file_path = "lakefs://example-repo/main~1/data.parquet"
```

### Benefits

1. **Improved Syntax Guarantee:** 
   The current approach involves extracting the commit ID programmatically:
   ```python
   prev_commit = repo.ref(f"{main.id}@").get_commit().id
   file_path = f"lakefs://example-repo/{prev_commit}/data.parquet"
   ```
   While this works, it introduces unnecessary complexity. Adding support for `@` or `~N` would streamline workflows and reduce boilerplate code.

2. **Consistency Across Tools:**
   Aligning `lakefs-spec` behavior with `lakectl` would create a unified and predictable user experience across lakeFS tools.

3. **Improved Usability:** 
   The syntactic sugar would significantly simplify code and make it easier for users to adopt and use lakeFS effectively.

### Proposed Implementation

- Update the URI parsing logic in `lakefs-spec` to recognize and correctly handle:
  - The `@` syntax for exact commit references.
  - The `~N` syntax for relative commit references.

- Ensure backward compatibility with existing paths.

### Example Usage After Implementation

With the updated functionality, users could directly specify:
```python
[...]

PR #316, opened 2025-01-16
Make lakeFS URI validation more lenient

AdrianoKF:
The URI validation logic was too strict in some cases, causing valid lakeFS ref expressions with relative (`~N`, `^N`) or HEAD (`@`) suffixes to be rejected.

This PR refactors the relevant regexp to accept these URIs and also makes the relative ref expression parsing more explicit in the regex.

Issue: #314

### DIFF
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -1,6 +1,8 @@
+import re
+
 import pytest
 
-from lakefs_spec.util import _batched
+from lakefs_spec.util import _batched, _uri_parts
 
 
 def test_batched_empty_iterable():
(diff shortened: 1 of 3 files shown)


tests/test_util.py
```diff
@@ -26,3 +28,44 @@ def test_batched_batch_size_greater_than_iterable():
...
+        "refexp, valid",
+        [
+            ("", False),
+            ("main", True),
+            ("main@", True),
+            ("main~", True),
+            ("main^", True),
+            ("main^2", True),
+            ("main^^^", True),
+            ("main^1^1", True),
```
nicholasjng, line 62: So `^N^N` is supported? Cool, did not know that, but it does make sense.

AdrianoKF: Note that I have not tried it against a lakeFS server, but the [docs](https://docs.lakefs.io/understand/model.html#ref-expressions) say that it should be supported:

> If `<ref>` is a ref expression, then:
> - [...]
> - `<ref>^N` is a ref expression referring to its N’th parent; in particular `<ref>^1` is the same as `<ref>^`.

Since this a recursive grammar, the repeated `^N` suffix should be accepted as a valid ref expression.

nicholasjng: Yeah, the caret `^` controls the ancestral depth, and the subsequent number gives the ancestor number. I think this works just like in git.

PR #318, opened 2025-02-28
refactor: Interpret all rpaths as resource names in transactions

nicholasjng:
Built on #317.

-------------------------------

Proof of concept for a single repo-and-branch-scoped transaction.

This ties a transaction to a single repo and branch by taking all file and
directory names by resource only instead of a full URI.

Naturally, this has the subtle side effect that given full URIs are silently
understood as nested paths, and uploaded to the transaction branch without
loud errors or warnings. A section was added to the transaction docs that
details this behavior, but it might be safer to check the input path against
existing repos and branches.

cc @AdrianoKF, after the UPathManager excursion we had lately. Opinions welcome - I'm leaning towards at least validating the input path against existing repos to make sure that the user doesn't accidentally do the wrong thing.

But in general, in my opinion, scoping transactions to a single repo and branch pair only makes sense, insofar as uploading resources to other repos and branches results in silent uncommitted changes, which is also not ideal.

Outstanding test failures are because I haven't ported all methods to the `make_uri(rpath)` pattern yet - I'm not sure this is the most elegant way to do things, so if you have other ideas, please let me know.

### DIFF
--- a/docs/guides/transactions.md
+++ b/docs/guides/transactions.md
@@ -20,9 +20,9 @@ from lakefs_spec import LakeFSFileSystem
 fs = LakeFSFileSystem()
 
 with fs.transaction("repo", "main") as tx:
-    fs.put_file("train-data.txt", f"repo/{tx.branch.id}/train-data.txt")
+    fs.put_file("train-data.txt", "train-data.txt")
     tx.commit(message="Add training data")
-    fs.put_file("test-data.txt", f"repo/{tx.branch.id}/test-data.txt")
+    fs.put_file("test-data.txt", "test-data.txt")
     sha = tx.commit(message="Add test data")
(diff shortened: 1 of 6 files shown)


nicholasjng:
An example that would be a silent failure case with this change:

```python

fs = LakeFSFileSystem()

with fs.transaction("my-repo", "my-branch"):
    fs.put_file("hello.txt", "my-other-repo/my-other-branch/hello.txt")  # <- creates the file my-repo/my-branch/my-other-repo/my-other-branch/hello.txt.
```

nicholasjng:
This is now ready for review. I'm not _extremely_ happy with how it turned out, but it does make the transaction a lot easier to use by not having to prepend repo and branch name all the time.

PR #325, opened 2025-07-16
Modify open() to use pre_sign storage configuration as default

DamienMatias:
The current `open()` implementation hardcodes `pre_sign=False` as the default. This PR changes it to `pre_sign=None` to use by default the value inferred from the storage configuration, as done by the underlying `lakefs` library.

### DIFF
--- a/tests/test_lakefs_file.py
+++ b/tests/test_lakefs_file.py
@@ -70,3 +70,21 @@ def test_lakefs_file_unknown_mode(fs: LakeFSFileSystem) -> None:
 
     with pytest.raises(NotImplementedError, match="unsupported mode .*"):
         fs.open("hello.py", mode="ab")  # type: ignore
+
+
+def test_lakefs_file_open_pre_sign_none_uses_storage_config(
+    fs: LakeFSFileSystem,
+    repository: Repository,
+    temp_branch: Branch,
+    random_file_factory: RandomFileFactory,
+) -> None:
+    """Test that pre_sign=None uses the storage configuration's pre_sign_support value."""
+    rpath = put_random_file_on_branch(random_file_factory, fs, repository, temp_branch)
+
+    # Open file with pre_sign=None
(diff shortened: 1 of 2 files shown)


tests/test_lakefs_file.py
```diff
@@ -70,3 +70,21 @@ def test_lakefs_file_unknown_mode(fs: LakeFSFileSystem) -> None:
...
+    temp_branch: Branch,
+    random_file_factory: RandomFileFactory,
+) -> None:
+    """Test that pre_sign=None uses the storage configuration's pre_sign_support value."""
+    rpath = put_random_file_on_branch(random_file_factory, fs, repository, temp_branch)
+
+    # Open file with pre_sign=None
+    with fs.open(rpath, mode="rb", pre_sign=None) as fp:
+        # Get the expected pre_sign value from storage config
+        expected_pre_sign = fp._client.storage_config_by_id(fp._obj.storage_id()).pre_sign_support
```
nicholasjng, line 87: This line fails on older versions of `lakefs`, see the actions logs.

Apart from that, this reads to me as if the storage config (and thus the value of `pre_sign_support`) can be set on the object level - could you clarify this? I thought that pre-signed URL support is a feature of the storage system, so it's either available for all objects or none of them.

If it really is object-specific, it might be worth updating the `open()` docstring to mention that fact.

nicholasjng: Found it - https://github.com/treeverse/lakeFS/commit/b87f996f92146179b4a9aa51f5ac031c8333e004 adds per-object storage configs.

For backwards compatibility, you could change to this:

```suggestion
        client = fp._client
        # Get the expected pre_sign value from storage config
        if hasattr(client, "storage_config_by_id"):
            expected_pre_sign = client.storage_config_by_id(fp._obj.storage_id()).pre_sign_support
        else:
            expected_pre_sign = client.storage_config.pre_sign_support
```

nicholasjng:
Thanks!

Issue #329, opened 2025-09-09
UPath function don't allow to pass a timeout to the underlying rest API calls


stephane-additive:
### What happened?

the fsspec allows most function to take an extra `kwargs` that could be passed to the udnerlying `objects_api.get_object` to allow setting a timeout to our lakefs request.

Background: We have lakefs reads that hang forever in urrlib. Here's the thread backtrace:
```
read (ssl.py:1103)
    recv_into (ssl.py:1251)
    readinto (socket.py:720)
    _safe_read (http/client.py:640)
    read (http/client.py:495)
    _fp_read (urllib3/response.py:887)
    _raw_read (urllib3/response.py:904)
    read (urllib3/response.py:980)
    __init__ (urllib3/response.py:664)
    getresponse (urllib3/connection.py:579)
    _make_request (urllib3/connectionpool.py:534)
    urlopen (urllib3/connectionpool.py:787)
    urlopen (urllib3/poolmanager.py:459)
    urlopen (urllib3/poolmanager.py:505)
    request_encode_url (urllib3/_request_methods.py:182)
    request (urllib3/_request_methods.py:135)
    request (lakefs_sdk/rest.py:206)
    get_request (lakefs_sdk/rest.py:243)
    request (lakefs_sdk/api_client.py:436)
    __call_api (lakefs_sdk/api_client.py:212)
    call_api (lakefs_sdk/api_client.py:410)
    get_object_with_http_info (lakefs_sdk/api/objects_api.py:738)
    execute (pydantic/v1/decorator.py:206)
    call (pydantic/v1/decorator.py:134)
    wrapper_function (pydantic/v1/decorator.py:40)
    get_object (lakefs_sdk/api/objects_api.py:605)
    execute (pydantic/v1/decorator.py:206)
    call (pydantic/v1/decorator.py:134)
    wrapper_function (pydantic/v1/decorator.py:40)
    _read (lakefs/object.py:288)
    read (lakefs/object.py:318)
```

### lakeFS-spec version

0.11.1

### lakeFS version

0.8.0

AdrianoKF:
[...]

PR #330, opened 2025-09-15
feat(fs): Pass kwargs to SDK object wrapper calls

nicholasjng:
Then, they are supposed to be forwarded to the raw lakeFS client API calls. This makes it possible to pass relevant parameters like timeouts to the actual API calls that are happening under the hood.

---------------------

Works towards #329. The PR on the lakeFS side, which is https://github.com/treeverse/lakeFS/pull/9509, provides kwargs support to many, but I think not all object functionality.

Needs testing against a new `lakefs` Python package release, and better docstrings.

### DIFF
--- a/tests/test_fs.py
+++ b/tests/test_fs.py
@@ -8,6 +8,7 @@ from packaging.version import Version
 from pytest import MonkeyPatch
 
 from lakefs_spec import LakeFSFileSystem
+from lakefs_spec.types import RequestConfig
 
 lakefs_version = Version(version("lakefs"))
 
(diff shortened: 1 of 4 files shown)


tests/test_fs.py
```diff
@@ -62,3 +71,26 @@ def test_initialization(monkeypatch: MonkeyPatch, temporary_lakectl_config: str)
     assert config.host == "http://lakefs.hello/api/v1"
     assert config.username == "my-user"
     assert config.password == "my-password"
+
+
+@pytest.mark.skipif(lakefs_version < Version("0.14"), reason="requires lakefs>=0.14.0")
+def test_request_config(
```
AdrianoKF, line 77: While this is a fine test case, I would also like to exercise the code path that passes the request args to the FS constructor to validate the logic that transforms the passed args.

PR #340, opened 2026-05-07
deps: Add uv dependency cooldown, improve zizmor config, update minimum fsspec version

nicholasjng:
Also add a uv config option to exclude releases newer than one week.


nicholasjng:
Failure is due to a minimum version requirement for fsspec in pandas:

```
FAILED tests/smoke_tests/test_integrations.py::test_pandas_integration - ImportError: Pandas requires version '2024.10.0' or newer of 'fsspec' (version '2023.12.0' currently installed).
```

I guess it's fine to bump?

PR #345, opened 2026-09-16
Switch to prek + ty in development, return UTC timestamps only 

nicholasjng:
Along with some dependency bumps and the mentioned switch to prek + ty (and locally sourced hooks for ruff and ty), this stack carries two breaking changes:

1) The `precheck` argument can now only be passed by keyword. Passing it positionally always produced wrong results, but now it's explicit.
2) The `created` and `modified` methods of the filesystem now return the timestamps in UTC time. This is done exactly as in fsspec, and silently produced times in the locale of the caller, which could lead to different answers on different machines.


nicholasjng:
I didn't yet raise the Python floor to 3.11, since Python 3.10 becomes EOL only on October 31st.

